PR #43, opened 2022-11-22
CM-16944 - Fix unique id concatenation in commit range scan

avishaiamiel:


tests/test_zip_file.py
```diff
@@ -0,0 +1,6 @@
+from cli import zip_file
+
+
+def test_concat_unique_id():
+    assert zip_file.concat_unique_id('path/to/file', 'unique_id') == 'unique_id/path/to/file'
```
MichalBor, line 5: please separate to two different tests

MichalBor (approved):

PR #48, opened 2022-12-06
CM-17071 - Support monitor param

avishaiamiel:


cli/code_scanner.py
```diff
@@ -31,20 +31,31 @@
...
     """ Scan git repository including its history """
     try:
         logger.debug('Starting repository scan process, %s', {'path': path, 'branch': branch})
+        scan_type = context.obj["scan_type"]
+        if monitor and scan_type != SCA_SCAN_TYPE:
+            raise click.ClickException(f"Monitor is currently supported for SCA scan type only")
+
         documents_to_scan = [
-            Document(get_path_by_os(os.path.join(path, obj.path)),
+            Document(obj.path if monitor else get_path_by_os(os.path.join(path, obj.path)),
```
MichalBor, line 51: why?

cli/code_scanner.py
```diff
@@ -31,20 +31,31 @@
...
+@click.option('--monitor',
+              is_flag=True,
+              default=False,
+              help="When specified, the scan results will be sent to Cycode platform and results will be parsed as "
+                   "vulnerabilities and violations.",
+              type=bool,
+              required=False)
 @click.pass_context
-def scan_repository(context: click.Context, path, branch):
+def scan_repository(context: click.Context, path, branch, monitor):
```
MichalBor, line 42: give it a boolean name

cli/code_scanner.py
avishaiamiel: the idea behind it is that monitor is something similar to push event. the detections should be opened over the path in the repository and not the path of the CLI machine.

cli/code_scanner.py
avishaiamiel: what is the different between this to all other flags that we have in code - verbose/show-secrets/soft-fail?

MichalBor (approved):

PR #49, opened 2022-12-08
CM-17670 - Fix Gradle dependencies location in monitor mode

avishaiamiel:


cli/helpers/sca_code_scanner.py
```diff
@@ -10,11 +11,14 @@
 BUILD_GRADLE_DEP_TREE_TIMEOUT = 180
 
 
-def run_pre_scan_actions(documents_to_scan: List[Document], is_git_diff: bool = False):
+def run_pre_scan_actions(context: click.Context, documents_to_scan: List[Document], is_git_diff: bool = False):
+    is_monitor = context.params.get('monitor')
+    project_path = context.params.get('path')
```
MichalBor, line 16: check if not null

cli/helpers/sca_code_scanner.py
```diff
@@ -10,11 +11,14 @@
 BUILD_GRADLE_DEP_TREE_TIMEOUT = 180
 
 
-def run_pre_scan_actions(documents_to_scan: List[Document], is_git_diff: bool = False):
+def run_pre_scan_actions(context: click.Context, documents_to_scan: List[Document], is_git_diff: bool = False):
+    is_monitor = context.params.get('monitor')
```
MichalBor, line 15: don't take from param, add it to context.obj in scan_repository

cli/helpers/sca_code_scanner.py
avishaiamiel: in all flows that we are calling it it's not null

cli/helpers/sca_code_scanner.py
```diff
@@ -12,13 +12,13 @@
 
 
 def run_pre_scan_actions(context: click.Context, documents_to_scan: List[Document], is_git_diff: bool = False):
-    is_monitor = context.params.get('monitor')
+    is_monitor_action = context.obj.get('monitor')
```
MichalBor, line 15: since its set only in repository scan, it can be null in case of path scan. 
please change to context.obj.get('monitor', false)

MichalBor (approved):

PR #57, opened 2023-02-03
CM-19861: Add SAST scan

sergey-rynkevich:


cli/code_scanner.py
```diff
@@ -775,6 +775,11 @@ def _map_detections_per_file(detections) -> List[DetectionsPerFile]:
     return [DetectionsPerFile(file_name=file_name, detections=file_detections)
             for file_name, file_detections in detections_per_files.items()]
 
+def _get_file_name_from_detection(detection):
+    if detection['source_policy_type'] == "SAST":
```
MichalBor, line 779: this is the old category field, please check `category` value. if you don't get it in the response so add it please

MichalBor (commented):
👏🏻

cli/code_scanner.py
sergey-rynkevich: it's not `category` - this is `policyType` field

MichalBor: `category` is the name of `policyTypeV2` on the detection entity

MichalBor (approved):

PR #74, opened 2023-03-02
CM-20322  Cycode CLI - report - support new option --report on Cycode CLI command

morsa4406:
…CLI command


cli/code_scanner.py
```diff
@@ -197,26 +225,34 @@ def pre_receive_scan(context: click.Context, ignored_args: List[str]):
...
     sca_code_scanner.perform_pre_hook_range_scan_actions(git_head_documents, pre_committed_documents)
     return scan_commit_range_documents(context, git_head_documents, pre_committed_documents,
-                                       configuration_manager.get_sca_pre_commit_timeout_in_seconds())
+                                       configuration_manager.get_sca_pre_commit_timeout_in_seconds(),
+                                       scan_parameters=scan_parameters)
 
 
 def scan_sca_commit_range(context: click.Context, path: str, commit_range: str):
+    report = context.obj["report"]
+    scan_parameters = get_scan_parameters(path, None, report)
```
MichalBor, line 241: why None, its boolean

MichalBor: since the only use of report is here and his use is just to pass it to another func, pass here the context and get the report value inside

MichalBor (changes requested):

cli/cycode.py
```diff
@@ -87,8 +87,14 @@
                    "vulnerabilities and violations (supported for SCA scan type only).",
               type=bool,
               required=False)
+@click.option('--report',
+              is_flag=True,
+              default=None,
```
MichalBor, line 92: False

MichalBor (changes requested):

MichalBor (approved):

PR #79, opened 2023-03-16
CM-21226 - fix commit range slowness

MichalBor:


MichalBor:
@doratias18 logs of?

doratias18:
@MichalBor progress, how many files, how many commits etc

PR #101, opened 2023-04-20
CM-22206 Add "auth check" command

MarshalX:


MarshalX:
rework of auth cmd planned in CM-22207

MarshalX:
@MichalBor I start thinking about the move `--output` option to global space. It should be `cycode --output json` for example. Otherwise, the option position will be dynamic and depend on the command space. There is `cycode scan --output`. And in this PR `cycode auth --output`. What do you think?

artem-fedorov (approved):

MichalBor:
> @MichalBor I start thinking about the move `--output` option to global space. It should be `cycode --output json` for example. Otherwise, the option position will be dynamic and depend on the command space. There is `cycode scan --output`. And in this PR `cycode auth --output`. What do you think?

@MarshalX sounds good

cli/auth/auth_command.py
```diff
@@ -18,6 +35,37 @@ def authenticate(context: click.Context):
         _handle_exception(context, e)
 
 
+@authenticate.command(name='check')
+@click.pass_context
+def authorization_check(context: click.Context):
+    """ Check your machine associating CLI with your cycode account """
+    def print_result(*, result: bool) -> None:
```
MichalBor, line 42: please do not create funcs inside other funcs.

cli/auth/auth_command.py
```diff
@@ -18,6 +35,37 @@ def authenticate(context: click.Context):
...
+        # the current impl of printers supports only results of scans
+        if context.obj['output'] == 'text':
+            secho_args = {
```
MichalBor, line 45: what about something like:
`failed_auth_check_args = {'message': 'You are not authorized', 'fg': 'red'}`
`passed_auth_check_args = {'message': 'You are not authorized', 'fg': 'red'}`
`auth_check_result = passed_auth_check_args if check_passed else failed_auth_check_args`
`return click.secho(**auth_check_result)`

cli/auth/auth_command.py
```diff
@@ -18,6 +35,37 @@ def authenticate(context: click.Context):
...
+    client_id, client_secret = CredentialsManager().get_credentials()
+    if not client_id or not client_secret:
+        return print_result(result=False)
+
+    try:
+        # TODO(MarshalX): This property performs HTTP request to refresh the token. This must be the method.
+        if CycodeTokenBasedClient(client_id, client_secret).api_token:
+            return print_result(result=True)
+    except (CycodeError, HTTPError):    # TODO(MarshalX): "HTTPError: 401" should be wrapped with our custom exception
+        if context.obj['verbose']:
```
MichalBor, line 63: nice!

cli/auth/auth_command.py
```diff
@@ -18,6 +35,37 @@ def authenticate(context: click.Context):
...
+
+    client_id, client_secret = CredentialsManager().get_credentials()
+    if not client_id or not client_secret:
+        return print_result(result=False)
+
+    try:
+        # TODO(MarshalX): This property performs HTTP request to refresh the token. This must be the method.
+        if CycodeTokenBasedClient(client_id, client_secret).api_token:
+            return print_result(result=True)
+    except (CycodeError, HTTPError):    # TODO(MarshalX): "HTTPError: 401" should be wrapped with our custom exception
```
MichalBor, line 62: regarding the TODO, please do it as part of this pr

MarshalX:
@MichalBor the option was moved to the global space. I saved backward comparability with `scan --output` option too. The cyclient was refactored and all error handling happens in the base class. Also I migrate a few direct calls of "requests" lib to cyclient

cli/auth/auth_command.py
```diff
@@ -1,14 +1,23 @@
...
 from cli.auth.auth_manager import AuthManager
-from cli.exceptions.custom_exceptions import AuthProcessError, CycodeError
+from cli.user_settings.credentials_manager import CredentialsManager
+from cli.exceptions.custom_exceptions import AuthProcessError, NetworkError, HttpUnauthorizedError
 from cyclient import logger
+from cyclient.cycode_token_based_client import CycodeTokenBasedClient
 
 
-@click.command()
+@click.group(invoke_without_command=True)
```
MichalBor, line 13: maybe there is a better way instead of set the command as a group?

cli/exceptions/custom_exceptions.py
```diff
@@ -18,17 +26,18 @@ def __str__(self):
         return f'error occurred during the scan. error message: {self.error_message}'
 
 
-class HttpUnauthorizedError(Exception):
-    def __init__(self, error_message: str):
+class HttpUnauthorizedError(CycodeError):   # TODO(MarshalX): maybe inhered from NetworkError?
```
MichalBor, line 29: I think its right to inherit from `CycodeError`
remove the TODO comment :)

cli/auth/auth_command.py
MarshalX: I did it to save backward comparability. It should look like our scan commands:
- scan path ...
- scan repository ...

the best-looking case:
- auth login
- auth check

but due to the already existing command `auth` (not the group of commands), we should do `invoke_without_command` trick to save prev usage of the command.

the other way to workaround it is to move the check command out of the "auth" space

for example: `cycode status` looks also great! and we can return many things on it (same with git status)! wdyt? @MichalBor

MarshalX: @artem-fedorov cc

MichalBor: @MarshalX thanks for the great explanation! 
1. lets continue with `cycode auth` and `cycode auth check` as planned, seems that the decorator you suggested is the best option at the moment :)
2. regarding `cycode status` - good idea! lets add it to the roadmap

MarshalX:
force pushed. re sign commit due to manipulation from GitHub side

cli/auth/auth_command.py
MarshalX: you are welcome :)

1. is pr approved?) 
2. added

cyclient/cycode_token_based_client.py
```diff
@@ -29,23 +29,23 @@ def refresh_api_token_if_needed(self) -> None:
...
-        self._expires_in = arrow.utcnow().shift(
-            seconds=auth_response_data['expires_in'] * 0.8)
+        self._expires_in = arrow.utcnow().shift(seconds=auth_response_data['expires_in'] * 0.8)
 
-    def get_request_headers(self, additional_headers: dict = None):
+    def get_request_headers(self, additional_headers: dict = None) -> dict:
         headers = super().get_request_headers(additional_headers=additional_headers)
-        headers = self._add_auth_header(headers)
+
+        if not self.lock.locked():
```
MichalBor, line 44: what was the race condition that it comes to solve?

MichalBor (commented):
small comments

cyclient/cycode_client_base.py
```diff
@@ -48,20 +47,39 @@ def _execute(
...
+            response.raise_for_status()
+            return response
+        except Exception as e:
+            self._handle_exception(e)
 
-    def get_request_headers(self, additional_headers: dict = None):
+    def get_request_headers(self, additional_headers: dict = None) -> dict:
         if additional_headers is None:
-            return self.MANDATORY_HEADERS
+            return self.MANDATORY_HEADERS.copy()
```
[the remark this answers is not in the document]
MarshalX: https://github.com/cycodehq-public/cycode-cli/pull/101/files#diff-c387cab8ff8e92fc0d3c1c614a6be5c543ba6633837d900aa72fb27a603bb695R45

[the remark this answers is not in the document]
MarshalX: it returned by reference. well, it's possible to easily mutate it by mistake.

cyclient/cycode_token_based_client.py
MarshalX: this solves the problem when we are attempting to fetch the token for the first time. when we request it we don't have an access token yet. so we need to perform the request without Auth header. without the lock here is the deadlock

tests/cli/test_cycode.py
```diff
@@ -26,14 +26,25 @@ def _is_json(plain: str) -> bool:
...
 
     responses.add(get_zipped_file_scan_response(get_zipped_file_scan_url(scan_type, scan_client)))
     responses.add(api_token_response)
     # scan report is not mocked. This raise connection error on attempt to report scan. it doesn't perform real request
 
-    args = ['scan', '--soft-fail', '--output', output, 'path', str(_PATH_TO_SCAN)]
+    args = ['scan', '--soft-fail', 'path', str(_PATH_TO_SCAN)]
+
+    if option_space == 'global':
+        args = [*['--output', output], *args]
```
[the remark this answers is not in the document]
MarshalX: Unfortunately the .extend() adds items to the end of args array. we need to add it to the global space of the command. to the start, to be more precise. that's why I did this trick(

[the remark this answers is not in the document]
MarshalX: well, I guess we can do ['--output', output].extend(args)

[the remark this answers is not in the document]
MichalBor: right, so maybe just `['--output', output] + args`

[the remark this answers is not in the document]
MarshalX: yeap, the snippet below is more readable. thank you! 

```python
global_args = ['--output', output]
global_args.extend(args)

args = global_args
```

MichalBor (approved):

PR #118, opened 2023-05-25
CM-23451 - Add "user-agent" global option to CLI

MarshalX:


cycode/cli/utils/machine_id.py
```diff
@@ -0,0 +1,115 @@
...
+import subprocess
+from sys import platform
+from typing import Optional
+
+from cycode.cli.exceptions.custom_exceptions import CycodeError
+
+
+def _read_cmd(cmd: str) -> Optional[str]:
+    try:
+        return subprocess.run(cmd, shell=True, capture_output=True, check=True, encoding='UTF-8').stdout.strip()
```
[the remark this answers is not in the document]
MarshalX: #cycode_sast_ignore_here

tests/utils/test_machine_id.py
```diff
@@ -0,0 +1,27 @@
...
+
+_APP_ID = 'TestsCycodeCLI'
+
+
+def test_hmac_hex():
+    key = 'ee8cd7bb-b427-4f56-ab0b-e02271cfce5c'
+    msg = 'e9913617-37ea-4ade-bbf9-58e3bf6c7937'
+    hmac_hex = _get_hmac_hex(key, msg)
+
+    expected_hmac_hex = '4dd8f70c07dd2b659c77a784169e9625960af8d598461a573522725a2be08f49'
```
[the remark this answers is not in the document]
MarshalX: #cycode_secret_ignore_here

MarshalX:
Ex. CMD:

```shell
poetry run cycode --user-agent='{"app_name": "vscode_extension", "app_version": "0.1.2", "env_name": "vscode", "env_version": "2023.1.2"}' scan path ...
```

Ex. User-Agent:
```text
CycodeCLI/0.0.0 (OS: Darwin; Arch: arm64; Python: 3.8.16; MID: 3f5c28fmyAwesomeHash80508dbb07ca1) vscode_extension (AppVersion: 0.1.2; EnvName: vscode; EnvVersion: 2023.1.2)
```

cycode/cli/utils/machine_id.py
```diff
@@ -0,0 +1,117 @@
...
+
+@lru_cache(maxsize=None)
+def machine_id() -> str:
+    return _get_machine_id()
+
+
+@lru_cache(maxsize=None)
+def protected_machine_id(app_id: str) -> str:
+    """Calculates HMAC-SHA256 of the app ID, keyed by the machine ID and returns a hex-encoded str."""
+    return _get_hmac_hex(key=_get_machine_id(), msg=app_id)
```
artem-fedorov, line 117: I don't really like collecting so much deep data and running shell here
I suggest to have the following:

"{OS_type}-{python_version}-{uniq_id}"
where uniq_id will be a hex/sha from one of the following options:
- mac address
- installation/home dir
- we'll write some uniq uuiq to `~/.cycode/id` file

cycode/cli/main.py
```diff
@@ -156,6 +164,10 @@ def main_cli(context: click.Context, verbose: bool, output: str):
 
     context.obj['output'] = output
 
+    if user_agent:
```
artem-fedorov, line 167: why not to send it in all queries?

cycode/cyclient/cycode_client_base.py
```diff
@@ -1,20 +1,46 @@
...
+    app_name = 'CycodeCLI'
+    version = __version__
+
+    os = platform.system()
+    arch = platform.machine()
+    python_version = platform.python_version()
+
+    mid = protected_machine_id(app_name)
+
+    return f'{app_name}/{version} (OS: {os}; Arch: {arch}; Python: {python_version}; MID: {mid})'
```
artem-fedorov, line 26: nice! see my comment above about `{mid}`

cycode/cli/utils/machine_id.py
MarshalX: Here is only one exec/file read. The implementation supports all environments and operation systems. That's why code contains many execs. The purpose of the code is simple: get OS installation ID. 

To get a mac address we should implement something similar. Because the way to retrieve it depends on the OS too. 

The solution with generating cycode cli installation ID instead of using OS installation ID sounds okay and was my second option to machine id

cycode/cli/main.py
MarshalX: Because here we enrich the user agent with provided information in the CLI option. The default user agent will be sent with each request.

Only if the user agent option is passed the base user agent will be enriched with application information that works with our CLI. I mean our plugins that use CLI under the hood

cycode/cli/utils/machine_id.py
artem-fedorov: do we cache somewhere info about the installation id in the machine?

MarshalX: Installation ID of OS? It cached in runtime. 

We don't have any logic for the installation ID of CLI yet

cycode/cyclient/cycode_client_base.py
```diff
@@ -1,20 +1,46 @@
...
-from cycode.cli.exceptions.custom_exceptions import NetworkError, HttpUnauthorizedError
+from ..cli.exceptions.custom_exceptions import NetworkError, HttpUnauthorizedError
+from ..cli.utils.machine_id import protected_machine_id
 
 
-class CycodeClientBase:
+def get_cli_user_agent() -> str:
+    """Return base User-Agent of CLI.
+
+    Example: CycodeCLI/0.2.3 (OS: Darwin; Arch: arm64; Python: 3.8.16; MID: *hash*)
```
MichalBor, line 15: nice!

cycode/cli/utils/machine_id.py
MichalBor: why we need the machine id? I agree with @artem-fedorov , lets keep it simple

artem-fedorov: we need some uniq id for build some dashboards how many users of CLI do we have :) 
I already solved similar issue in jmeter-plugins : https://github.com/undera/jmeter-plugins-manager/blob/master/src/main/java/org/jmeterplugins/repository/JARSourceHTTP.java#L304

MichalBor: @MarshalX field in config.yaml sounds good to me. @artem-fedorov WDYT?

artem-fedorov: who generate `config.yaml` and how often do we refresh it?

MichalBor: we generate it if its needed, for exmaple for adding ignores by value/secret/etc. it can be generated somehow in the first run for adding the id. also we can get another id of the local comp, but not that id that requires lots of logic haha :)

MarshalX: > who generate `config.yaml` and how often do we refresh it?

CLI creates if not exist. Refresh? What do you mean. We don't delete this config at all

artem-fedorov: In case this is our internal id, we can save random uuiq in this file

MarshalX:
this is how it looks in the confg.yaml 

image

artem-fedorov (approved):

artem-fedorov (approved):

PR #119, opened 2023-05-28
CM-23340 - Skip on collect node_modules folder for SCA npm 

PelegCycode:


MarshalX:
@PelegCycode Could you please test your `_is_file_relevant_for_sca_scan` function directly by faking path strings using "os.path" module? I mean without creating files with some content like now

This is example how we can test path-related functions that will work on all operation system:
https://github.com/cycodehq-public/cycode-cli/blob/d071284556c2ae139480152e72884b1701c5f658/tests/test_zip_file.py#L16-L22

as you can see `concat_unique_id` works with strings too

MarshalX:
@PelegCycode the test looks great! Could you pls delete the prev one? and tests JS files that were commited

MarshalX (approved):

PR #122, opened 2023-06-09
CM-23720 - Add table output for all scan types

MarshalX:


cycode/cli/printers/table_printer.py
```diff
@@ -1,170 +1,168 @@
...
-PREVIEW_DETECTIONS_COMMON_HEADERS = [
-    'File Path',
-    'Ecosystem',
-    'Dependency Name',
-    'Direct Dependency',
-    'Development Dependency'
+
+class ColumnInfo(NamedTuple):
+    name: str
+    width_secret: int = 1
```
MichalBor, line 15: not sure I understand whats the purpose of this class

cycode/cli/printers/table_printer.py
```diff
@@ -1,170 +1,168 @@
...
+        if self.scan_type == SECRET_SCAN_TYPE:
+            issue_type = detection.type
 
-    def _print_table_detections(
-            self, detections: List[Detection], headers: List[str], rows, title: str
-    ) -> None:
-        self._print_summary_issues(detections, title)
-        text_table = Texttable()
-        text_table.header(headers)
+        rows = [
```
MichalBor, line 124: detection columns values, no?

cycode/cli/printers/table_printer.py
```diff
@@ -1,170 +1,168 @@
...
+            issue_type,
+            detection.detection_rule_id,
+            click.format_filename(document_path),
+        ]
 
-        self.set_table_width(headers, text_table)
+        if self.scan_type == SECRET_SCAN_TYPE:
```
MichalBor, line 130: you already checked if secret scan type in line 121, unite the ifs :)

cycode/cli/printers/table_printer.py
```diff
@@ -1,170 +1,168 @@
...
+        rows = [
+            detection_line,
+            detection_position,
         ]
 
-        upgrade = ''
-        if alert.get("first_patched_version"):
-            upgrade = f'{alert.get("vulnerable_requirements")} -> {alert.get("first_patched_version")}'
-        row.append(upgrade)
+        if self.scan_type == SECRET_SCAN_TYPE:
```
MichalBor, line 153: you already check type above? its because the order of the columns? can we crate a map column<>value?

cycode/cli/printers/table_printer.py
```diff
@@ -1,170 +1,168 @@
...
+            rows.append(f'{violation_length} chars')
+
+            violation = ''
+
+            file_content_lines = document.content.splitlines()
+            if detection_line < len(file_content_lines):
```
MichalBor, line 159: from the results you attach in slack it seems you print only the line of the secret?

cycode/cli/printers/sca_table_printer.py
```diff
@@ -0,0 +1,166 @@
...
+PREVIEW_DETECTIONS_COMMON_HEADERS = [
+    'File Path',
+    'Ecosystem',
+    'Dependency Name',
+    'Direct Dependency',
+    'Development Dependency'
+]
+
+
+class SCATablePrinter(BasePrinter):
```
MichalBor, line 26: can we inherit from TablePrinter? it makes sense?

cycode/cli/printers/table_printer.py
MarshalX: To manage different widths of columns depends on scan type. For example in secret scan we display pretty short issue type. For example regex. But in sast we describe the issue. That's why I want to control columns widths for different scan types. This class allows to manage it

cycode/cli/printers/table_printer.py
MarshalX: Do you mean var name? Yeah, could be better. I just did in in the same style as the previous table printer for sca

cycode/cli/printers/sca_table_printer.py
MarshalX: IMHO the methods are so unique per table. but yeah, we could create BaseTablePrinter and move the constructor, print_result, and print_error methods (at least)

cycode/cli/printers/table_printer.py
MarshalX: you are right. tbh i have no clue how to make it pretty. i mean how to show the full line or few lines of the code in the small cell of the table. for example for sast scan results. that's why i just show only hardcoded secret for now

cycode/cli/printers/table_printer.py
MarshalX: yeah. that's for the order

cycode/cli/printers/table_printer.py
MarshalX: that's also because of columns order :(

cycode/cli/printers/base_table_printer.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def get_columns_info(self) -> List[ColumnInfo]:
+        return self._get_ordered_columns()
+
+    def get_headers(self) -> List[str]:
+        return [header.name for header in self._get_ordered_columns()]
+
+    def get_rows(self) -> List[str]:
+        ordered_values = [self._columns[column_info] for column_info in self._get_ordered_columns()]
+        return list(zip(*ordered_values))
```
MichalBor, line 51: all the lists have the same length? if not we might omit lines when using zip. for example
[1,2,3] , [a,b,c], [!,@] -> [(1,a,!), (2,b,@)

cycode/cli/printers/base_table_printer.py
```diff
@@ -0,0 +1,112 @@
...
+        return sorted(self._columns, key=lambda column_info: column_info.index)
+
+    def get_columns_info(self) -> List[ColumnInfo]:
+        return self._get_ordered_columns()
+
+    def get_headers(self) -> List[str]:
+        return [header.name for header in self._get_ordered_columns()]
+
+    def get_rows(self) -> List[str]:
+        ordered_values = [self._columns[column_info] for column_info in self._get_ordered_columns()]
```
MichalBor, line 50: rename `ordered_values` to `column_values`

cycode/cli/printers/base_table_printer.py
```diff
@@ -0,0 +1,112 @@
...
+
+from cycode.cli.consts import SECRET_SCAN_TYPE, INFRA_CONFIGURATION_SCAN_TYPE, SAST_SCAN_TYPE
+from cycode.cli.printers.text_printer import TextPrinter
+from cycode.cli.models import DocumentDetections, CliError, CliResult
+from cycode.cli.printers.base_printer import BasePrinter
+
+
+class ColumnInfo(NamedTuple):
+    name: str
+    index: int  # lower index means left column
```
MichalBor, line 15: Represents the order of the columns, starting from the left

cycode/cli/printers/base_table_printer.py
```diff
@@ -0,0 +1,112 @@
...
+    width_secret: int = 1
+    width_sast: int = 1
+    width_iac: int = 1
```
MichalBor, line 18: I think that its better to remove the mention of the scan types
the tuple should be (name,index,width)
calculate the width by the scan type before creating the ColumnInfo
you can remove also the _get_table_columns_width func that has mentions to scan types as well

MichalBor (commented):
I very like the new table management 💯

cycode/cli/printers/base_table_printer.py
MarshalX: Yeap. If not textable also will raise the exception. Blank cell means empty sting means array item

MichalBor (approved):

MaorDavidzon:
Can you please add a screenshot of the new table for each scan type? @MarshalX

cycode/cli/printers/table_printer.py
```diff
@@ -1,170 +1,116 @@
...
+        RULE_ID_COLUMN: 2,
+        FILE_PATH_COLUMN: 3,
+    },
+}
+
+
+class TablePrinter(BaseTablePrinter):
+    def _print_results(self, results: List[DocumentDetections]) -> None:
+        table = self._get_table()
+        if self.scan_type in COLUMN_WIDTHS_CONFIG:
```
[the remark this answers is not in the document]
MichalBor: I meant what happening if you don't have a key for the `self.scan_type`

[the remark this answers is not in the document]
MarshalX: this if statement check the scan_type key in the dict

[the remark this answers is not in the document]
MichalBor: I know, but if not you don't set the width but continue as usual and enriching the table. is it possible scenario? you validate the scan type in earlier stage?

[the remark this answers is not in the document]
MarshalX: if will fallback to default width by textable :) 

image

MarshalX:
> Can you please add a screenshot of the new table for each scan type? @MarshalX

Of course! 

### Secrets scan: 
image

### IAC scan
image

### SAST scan
image

PR #127, opened 2023-06-27
CM-24634 [SCA-CLI] adding dependency paths column on printing table result

morsa4406:


MarshalX:
@morsa4406 pls verify your task after refactoring. Pls note that we supported Python 3.7. it's illegal to use |. Use Union instead. | was added in Python 3.10

morsa4406:
@MarshalX Yes after refactoring, all tests has passed 🥇

MarshalX (approved):

PR #130, opened 2023-06-27
CM-22319 - Run secrets scanning asynchronously using polling mechanism

MarshalX:


doratias18:
@MarshalX still relevant? i think you did it in #129  ?

MarshalX:
> @MarshalX still relevant? i think you did it in #129 ?

it's still relevant. 129 about running many scans at the same time. this task is about moving from blocking scan requests to async requests using polling

PR #131, opened 2023-06-28
CM-24707 - Fix dependency paths to add continuation dependencies sign for the list greater than 2

morsa4406:


cycode/cli/printers/sca_table_printer.py
```diff
@@ -112,7 +112,7 @@ def _print_summary_issues(detections: List, title: str) -> None:
     def _shortcut_dependency_paths(dependency_paths: str) -> str:
         dependencies = dependency_paths.split(' -> ')
 
-        if len(dependencies) < 2:
+        if len(dependencies) < 3:
```
artem-fedorov, line 115: what is the magic value?

MarshalX (approved):

cycode/cli/printers/sca_table_printer.py
morsa4406: The dependency path should be mention only direct dependency until the problematic dependency.
If we have lower than 3, no changes should be on dependency paths, otherwise we need to remove all dependencies between direct dependencies to problematic dependency

cycode/cli/utils/string_utils.py
```diff
@@ -47,3 +47,13 @@ def generate_random_string(string_len: int):
 
 def get_position_in_line(text: str, position: int) -> int:
     return position - text.rfind('\n', 0, position) - 1
+
+
+@staticmethod
```
MarshalX, line 52: it is not a class member anymore. we should remove "@staticmethod"

morsa4406: Fixed

cycode/cli/printers/sca_table_printer.py
artem-fedorov: why do you need to remove path to problematic dependency?

morsa4406: all the middle dependencies between the direct dependency to vulnerable dependency since this path could be very long

artem-fedorov: it's a table output, right?
why not move dependency to a new line?
e.g. `A` or `B` option instead of `C`
image

morsa4406: It's also good option but there is list of dependencies paths and each has much more chars on dependency name which isn't readability

artem-fedorov: ok
go ahead with your approach, but
- move magic 3 to const
- use `<=` instead
- add documentation comment why do you have it the method

morsa4406: The result is:
a -> ... -> c

d -> ... -> f

morsa4406: @artem-fedorov fixed

artem-fedorov (approved):

PR #136, opened 2023-07-06
CM-22319 - Run secrets scanning asynchronously using polling mechanism

MarshalX:


cycode/cli/code_scanner.py
```diff
@@ -546,13 +546,13 @@ def perform_scan(
     is_commit_range: bool,
     scan_parameters: dict,
 ) -> ZippedFileScanResult:
-    if scan_type in (consts.SCA_SCAN_TYPE, consts.SAST_SCAN_TYPE):
-        return perform_scan_async(cycode_client, zipped_documents, scan_type, scan_parameters)
-
```
MaorDavidzon, line 551: I want to release it with a beta flag. for example, `cycode scan path --report-beta...`
Because I am not sure that the Async scan is stable yet on the backend side.

MaorDavidzon (changes requested):

cycode/cli/code_scanner.py
MarshalX: how and when we will figure out that it is stable? who will use this new flag?

MaorDavidzon: I will direct the customers to use the beta explicitly. 
The end-2-end does not test in high load and big repositories.

MarshalX: @MaorDavidzon customers could simply upgrade CLI to get async secrets scanning and downgrade it back to sync scans whenever (all historical releases are available on PyPI and could be installed with 1 command). Are you sure that we need to have this flag in the environment where it's possible to downgrade the CLI version and back to the stable one. tbh i rly don't want to start an era of feature flags in CLI

MaorDavidzon: Ok it's fine. But please don't release a new version without my approval

MarshalX:
i ran our e2e test locally with this PR. all passed

artem-fedorov (approved):

MarshalX:
@MaorDavidzon hi, can you approve it to merge to the main branch pls?

MaorDavidzon (approved):

PR #150, opened 2023-07-27
CM-25728 - Fix relative paths; fix validations of paths

MarshalX:


artem-fedorov (approved):

MarshalX:
before
image

after 
image

plus fix of relative paths

PR #154, opened 2023-08-09
CM-26097 - added disable restore flag to cli

PelegCycode:


cycode/cli/code_scanner.py
```diff
@@ -810,6 +811,7 @@ def get_git_repository_tree_file_entries(
 
 def get_default_scan_parameters(context: click.Context) -> dict:
     return {
+        SCA_DISABLE_RESTORE_DEPENDENCIES_FLAG: context.obj.get(SCA_DISABLE_RESTORE_DEPENDENCIES_FLAG),
         'monitor': context.obj.get('monitor'),
```
MarshalX, line 815: i think we should do it in the same style as is was. without the constant. could be refactored later for **all** options

cycode/cli/consts.py
```diff
@@ -189,3 +189,5 @@
 # Example: A -> B -> C
 # Result: A -> ... -> C
 SCA_SHORTCUT_DEPENDENCY_PATHS = 2
+
+SCA_DISABLE_RESTORE_DEPENDENCIES_FLAG = 'disable-restore-dependencies'
```
MarshalX, line 193: looks so complecated to type for users. should we at least add shortcut? 

we have -o for --option
-t for --scan-type 

and etc

cycode/cli/code_scanner.py
PelegCycode: it's cause our product hasn't yet decided on the flag name - it's much easier to change it with constant... I think it's better for most cases

cycode/cli/consts.py
PelegCycode: I'll consult with my product

cycode/cli/consts.py
```diff
@@ -189,3 +189,5 @@
 # Example: A -> B -> C
 # Result: A -> ... -> C
 SCA_SHORTCUT_DEPENDENCY_PATHS = 2
+
+SCA_DISABLE_RESTORE_DEPENDENCIES_FLAG = 'no-restore'
```
morsa4406, line 193: Please insert sca-no-restore

cycode/cli/main.py
```diff
@@ -111,6 +125,7 @@ def code_scan(
     sca_scan: List[str],
     monitor: bool,
     report: bool,
+    no_restore: bool,
```
morsa4406, line 128: sca_no_restore

cycode/cli/main.py
```diff
@@ -99,6 +105,14 @@
     type=bool,
     required=False,
 )
+@click.option(
+    '--' + SCA_DISABLE_RESTORE_DEPENDENCIES_FLAG,
+    is_flag=True,
+    default=False,
+    help='When declared, cycode will not run restore command. Will scan direct dependencies ONLY!',
```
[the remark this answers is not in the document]
MarshalX:
```suggestion
    help='When declared, Cycode will not run restore command. Will scan direct dependencies ONLY!',
```

maybe we should provide help text in the same style as `--no-progress-meter` flag? 

> Do not show the progress meter.

smth like "Do not run restore command.". wdyt?

cycode/cli/main.py
PelegCycode: flag name must be equal to parameters name

cycode/cli/consts.py
PelegCycode: the name of the flag has been reviewed with product

MarshalX: @morsa4406 we already have SCA only options without SCA prefix

like this:
>  `--severity-threshold [INFO\|LOW\|MEDIUM\|HIGH\|CRITICAL]`  | Show only violations at the specified level or higher (supported for the SCA scan type only).

WDYT?

MarshalX (approved):
LGTM! if the help text is approved feel free to merge

morsa4406 (approved):

niramna (changes requested):

MarshalX (approved):

niramna (approved):

PR #160, opened 2023-08-28
CM-26660-fix-typo-in-filename

idan-cyber:


idan-cyber:
Turns out both license and licence are correct.

PR #164, opened 2023-09-21
CM-24498 - Add SBOM reports

MarshalX:
TODO:
- [x] add progress bar support
- [x] add y/n question about already existing output file 
- [x] add error handling with user-friendly wrapping (like in scans)
- [x] add resolving of output file extension based on formats
- [x] sync available output formats with BE
- [x] sync available report formats with BE
- [x] update response models, cleanup communication with server after changes from BE team
- [x] **UPD bade UX i dont like it**. add not only overwrite YES/NO answers, but also SAVE BOTH
- [x] add metrics endpoint
- [x] add prompt to gen unique suffix for filename if already exists
- [x] update feedback endpoint after BE changes (some fields should be optional) 

DEMO

CLI interface:

image

File overwrite (files from 0-4 already exist):

image


MarshalX:
@PelegCycode @morsa4406 can't pick you as reviewers (you don't appear in the list). Feel free to comment too

PR #196, opened 2024-01-15
CM-31321 - Fix multiprocessing module in packaged CLI with Python 3.12 on macOS

MarshalX:


MichalBor (approved):
interesting!

MarshalX:
> interesting! the cli is working as expected when running the code not as executable?

yep it works fine

PR #197, opened 2024-01-17
CM-31412 - Fix severity sorting for unknown severity

morsa4406:


MarshalX:
as i understand it breaks this code:

https://github.com/cycodehq/cycode-cli/blob/7978be2c8bb91681510a95dca922a496a11b714e/cycode/cli/commands/scan/code_scanner.py#L756-L761

MarshalX:
ig we can create Severity.get_value (without try) and return -2 if non exists. and leave existing method without changes

or even get_sort_value/get_member_weight to be more clear. and this method will be used in table printer

morsa4406:
> as i understand it breaks this code:
> 
> https://github.com/cycodehq/cycode-cli/blob/7978be2c8bb91681510a95dca922a496a11b714e/cycode/cli/commands/scan/code_scanner.py#L756-L761

Fixed

MarshalX:
> ig we can create Severity.get_value (without try) and return -2 if not exists. and leave the existing method without changes
> 
> or even get_sort_value/get_member_weight to be more clear. and this method will be used in table printer

@morsa4406 in case you missed this. btw linter failed

MarshalX (approved):

PR #198, opened 2024-01-22
CM-31407 - Add new fields to scan parameters

MarshalX:


cycode/cli/commands/scan/code_scanner.py
```diff
@@ -570,12 +575,19 @@ def get_default_scan_parameters(context: click.Context) -> dict:
...
 
 
 def get_scan_parameters(context: click.Context, paths: Tuple[str]) -> dict:
     scan_parameters = get_default_scan_parameters(context)
 
+    if not paths:
+        return scan_parameters
+
+    scan_parameters['paths'] = paths
+    scan_parameters['path'] = paths[0]
```
MarshalX, line 589: @artem-fedorov i rly wanna send all paths to be able to switch logic on BE side. for now you could use the first one. in case if you dont want to update BE i added "path" param too. lemme know you you are okay to update BE and im gonna remove additional param

artem-fedorov: no need `path`

artem-fedorov (approved):

PR #203, opened 2024-01-24
CM-30564 - Add support for report command in secret scanning

saramontif:


cycode/cli/commands/scan/code_scanner.py
```diff
@@ -337,6 +348,7 @@ def scan_commit_range_documents(
                 scan_type,
                 scan_parameters,
                 timeout,
+                should_use_scan_service,
```
MarshalX, line 351: as i can see the `scan_commit_range_documents` method is used only for SCA (called in `scan_sca_pre_commit` and `scan_sca_commit_range`. SCA always goes to the scan service. i think we don't need here `should_use_scan_service` at all

for Secrets commit history scans `scan_documents` is called with `is_git_diff=True, is_commit_range=True` flags

cycode/cli/commands/scan/code_scanner.py
```diff
@@ -439,12 +452,24 @@ def perform_scan(
...
+    return poll_scan_results(
+        cycode_client,
+        scan_async_result.scan_id,
+        scan_type,
+        scan_parameters.get('report'),
+        should_use_scan_service=should_use_scan_service,
+    )
 
 
 def perform_commit_range_scan_async(
```
MarshalX, line 475: async scans in terms of code naming here are scans that go to scan service. see the prev comment for more thoughts. imo this method should not be changed

cycode/cli/commands/scan/code_scanner.py
```diff
@@ -575,7 +610,6 @@ def get_default_scan_parameters(context: click.Context) -> dict:
         'report': context.obj.get('report'),
         'package_vulnerabilities': context.obj.get('package-vulnerabilities'),
         'license_compliance': context.obj.get('license-compliance'),
-        'command_type': context.info_name,
```
MarshalX, line 578: pls return this line. probably missed during merging

cycode/cli/commands/scan/code_scanner.py
```diff
@@ -476,14 +511,14 @@ def poll_scan_results(
     end_polling_time = time.time() + polling_timeout
 
     while time.time() < end_polling_time:
-        scan_details = cycode_client.get_scan_details(scan_type, scan_id)
+        scan_details = cycode_client.get_scan_details(scan_type, scan_id, should_use_scan_service)
```
MarshalX, line 514: `get_scan_details` is used only in polling. polling only belongs to the scan service (AKA async scans). i think we don't need additional flags here

cycode/cli/commands/scan/code_scanner.py
```diff
@@ -769,37 +804,49 @@ def _does_severity_match_severity_threshold(severity: str, severity_threshold: s
...
-def init_default_scan_result(scan_id: str, scan_metadata: Optional[str] = None) -> ZippedFileScanResult:
+def init_default_scan_result(
+    cycode_client: 'ScanClient', scan_id: str, scan_type: str, should_get_report: bool = False
+) -> ZippedFileScanResult:
     return ZippedFileScanResult(
-        did_detect=False, detections_per_file=[], scan_id=scan_id, report_url=_try_get_report_url(scan_metadata)
+        did_detect=False,
+        detections_per_file=[],
+        scan_id=scan_id,
+        report_url=_try_get_report_url_if_needed(cycode_client, should_get_report, scan_id, scan_type),
```
MarshalX, line 835: what about something like 

```python
report_url = None
if should_get_report:
    report_url = _try_get_report_url()
```

instead of `_try_get_report_url_if_needed`

cycode/cyclient/models.py
```diff
@@ -171,6 +171,23 @@ def __init__(
         self.err = err
 
 
+class ScanReportUrlResponse(Schema):
```
MarshalX, line 174: as i understand this is DTO class and should not be inhered from Schema. you already have schema below

cycode/cyclient/models.py
```diff
@@ -171,6 +171,23 @@ def __init__(
...
+    def __init__(
+        self,
+        report_url: str,
+    ) -> None:
+        super().__init__()
```
MarshalX, line 179: pls use `@dataclass` instead. this is legacy approach. see the examples at the end of `models.py` file

cycode/cyclient/models.py
```diff
@@ -171,6 +171,23 @@ def __init__(
...
+        self.report_url = report_url
+
+
+class ScanReportUrlResponseSchema(Schema):
+    report_url = fields
```
MarshalX, line 184: ig you mean `fields.String()` instead of `fields`

cycode/cyclient/scan_client.py
```diff
@@ -72,17 +72,27 @@ def zipped_file_scan(
...
+        zip_file: InMemoryZip,
+        scan_type: str,
+        scan_parameters: dict,
+        is_git_diff: bool = False,
+        should_use_scan_service: bool = False,
     ) -> models.ScanInitializationResponse:
         files = {'file': ('multiple_files_scan.zip', zip_file.read())}
         response = self.scan_cycode_client.post(
-            url_path=self.get_zipped_file_scan_async_url_path(scan_type),
+            url_path=self.get_zipped_file_scan_async_url_path(scan_type, should_use_scan_service),
```
MarshalX, line 95: see the comment above

cycode/cyclient/scan_client.py
```diff
@@ -72,17 +72,27 @@ def zipped_file_scan(
 
         return self.parse_zipped_file_scan_response(response)
 
-    def get_zipped_file_scan_async_url_path(self, scan_type: str) -> str:
+    def get_scan_report_url(self, scan_id: str, scan_type: str) -> models.ScanReportUrlResponse:
+        response = self.scan_cycode_client.get(url_path=self.get_scan_report_url_path(scan_id, scan_type))
+        return models.ScanReportUrlResponseSchema().build_dto(response.json())
+
+    def get_zipped_file_scan_async_url_path(self, scan_type: str, should_use_scan_service: bool = False) -> str:
```
MarshalX, line 79: - `get_zipped_file_scan_async_url_path` method is used only in `zipped_file_scan_async`
- `zipped_file_scan_async`  method is used only in `perform_scan_async`

`perform_scan_async` performs scans via scan service only. no need for `should_use_scan_service` here

cycode/cyclient/scan_client.py
```diff
@@ -108,11 +118,16 @@ def multiple_zipped_file_scan_async(
         )
         return models.ScanInitializationResponseSchema().load(response.json())
 
-    def get_scan_details_path(self, scan_type: str, scan_id: str) -> str:
-        return f'{self.get_scan_service_url_path(scan_type)}/{scan_id}'
+    def get_scan_details_path(self, scan_type: str, scan_id: str, should_use_scan_service: bool = False) -> str:
```
MarshalX, line 121: it looks that `get_scan_details_path` is used only in `get_scan_details` which is used in `poll_scan_results`. and the polling is async scan (via scan service only). no need of `should_use_scan_service` here

cycode/cyclient/scan_client.py
```diff
@@ -108,11 +118,16 @@ def multiple_zipped_file_scan_async(
...
+
+    def get_scan_report_url_path(self, scan_id: str, scan_type: str) -> str:
+        return f'{self.get_scan_service_url_path(scan_type, True)}/reportUrl/{scan_id}'
 
-    def get_scan_details(self, scan_type: str, scan_id: str) -> models.ScanDetailsResponse:
-        path = self.get_scan_details_path(scan_type, scan_id)
+    def get_scan_details(
+        self, scan_type: str, scan_id: str, should_use_scan_service: bool = False
+    ) -> models.ScanDetailsResponse:
+        path = self.get_scan_details_path(scan_type, scan_id, should_use_scan_service)
```
MarshalX, line 130: see the comment above

cycode/cyclient/scan_config_base.py
```diff
@@ -38,9 +48,14 @@ def get_service_name(self, scan_type: str) -> str:
...
 
+    def get_scan_service_prefix(self) -> str:
+        return '5004'
+
 
 class DefaultScanConfig(ScanConfigBase):
-    def get_service_name(self, scan_type: str) -> str:
+    def get_service_name(self, scan_type: str, should_use_scan_service: bool = False) -> str:
+        if should_use_scan_service:
+            return self.get_scan_service_prefix()
```
MarshalX, line 58: if we want to have `get_scan_service_prefix` instead of literal pls replace the literal below (for sca and sast)

but imo as we already have literals for secrets and iac detector ig we are fine without additional abstract method here 

i mean it's duplication in both cases. for dev and default configs. so it could be more consistent without additional method

tests/cyclient/mocked_responses/scan_client.py
```diff
@@ -79,6 +79,20 @@ def get_scan_details_url(scan_id: Optional[UUID], scan_client: ScanClient) -> st
...
+def get_scan_report_url(scan_id: Optional[UUID], scan_client: ScanClient, scan_type: str) -> str:
+    api_url = scan_client.scan_cycode_client.api_url
+    service_url = scan_client.get_scan_report_url_path(str(scan_id), scan_type)
+    return f'{api_url}/{service_url}'
+
+
+def get_scan_report_url_response(url: str, scan_id: Optional[UUID] = None) -> responses.Response:
+    if not scan_id:
+        scan_id = uuid4()
+    json_response = {'report_url': 'https://app.domain/on-demand-scans/{scan_id}'.format(scan_id=scan_id)}
```
MarshalX, line 91: pls use f-strings instead of `.format`

tests/cyclient/scan_config/test_dev_scan_config.py
```diff
@@ -8,9 +8,16 @@ def test_get_service_name() -> None:
     assert dev_scan_config.get_service_name('iac') == '5026'
     assert dev_scan_config.get_service_name('sca') == '5004'
     assert dev_scan_config.get_service_name('sast') == '5004'
+    assert dev_scan_config.get_service_name('secret', True) == '5004'
```
MarshalX, line 11: please use a named argument for bool because it's not clear what it mean

instead of `dev_scan_config.get_service_name('secret', True)` use
`dev_scan_config.get_service_name('secret', should_use_scan_service=True)`

cycode/cli/commands/scan/code_scanner.py
```diff
@@ -148,14 +151,21 @@ def _scan_batch_thread_func(batch: List[Document]) -> Tuple[str, CliError, Local
...
         scan_id = str(_generate_unique_id())
         scan_completed = False
+        should_use_scan_service = _should_use_scan_service(scan_type, scan_parameters)
 
         try:
             logger.debug('Preparing local files, %s', {'batch_size': len(batch)})
             zipped_documents = zip_documents(scan_type, batch)
             zip_file_size = zipped_documents.size
-
             scan_result = perform_scan(
```
MarshalX, line 160: ig it is possible to move

`should_use_scan_service = _should_use_scan_service(scan_type, scan_parameters)`

inside of `perform_scan` function without additional args (because it's already too much)

and for debug logs `scan_type` and `scan_parameters` are enough to understand what is going on

cycode/cyclient/scan_client.py
```diff
@@ -108,11 +118,16 @@ def multiple_zipped_file_scan_async(
         )
         return models.ScanInitializationResponseSchema().load(response.json())
 
-    def get_scan_details_path(self, scan_type: str, scan_id: str) -> str:
-        return f'{self.get_scan_service_url_path(scan_type)}/{scan_id}'
+    def get_scan_details_path(self, scan_type: str, scan_id: str, should_use_scan_service: bool = False) -> str:
+        return f'{self.get_scan_service_url_path(scan_type, should_use_scan_service)}/{scan_id}'
+
+    def get_scan_report_url_path(self, scan_id: str, scan_type: str) -> str:
+        return f'{self.get_scan_service_url_path(scan_type, True)}/reportUrl/{scan_id}'
```
MarshalX, line 125: pls use named arg

`...scan_type, should_use_scan_service=True)`

cycode/cli/commands/scan/code_scanner.py
```diff
@@ -769,37 +804,49 @@ def _does_severity_match_severity_threshold(severity: str, severity_threshold: s
...
+) -> Optional[str]:
+    if not should_get_report:
         return None
 
     try:
-        metadata_json = json.loads(metadata_json)
-        return metadata_json.get('report_url')
-    except json.JSONDecodeError:
-        return None
+        report_url_response = cycode_client.get_scan_report_url(scan_id, scan_type)
```
MarshalX, line 846: is it compatible with SCA?

MarshalX (commented):
good job! pls see the comments below

cycode/cli/commands/scan/code_scanner.py
saramontif: We need this flag to get the correct service path

MarshalX: ah ( it getting more complex and complex 

 let's not use `get_scan_service_url_path` in `get_scan_details_path` at all. to not pass every time `should_use_scan_service`. wdyt?

cycode/cli/commands/scan/code_scanner.py
saramontif: Yes

MarshalX (approved):
well done! now it much much better and passes my small local runs

PR #207, opened 2024-01-25
CM-30564 - Print report URLs at the end of the table instead of the column

saramontif:


cycode/cli/printers/tables/table_printer.py
```diff
@@ -59,13 +59,12 @@ def _print_results(self, local_scan_results: List['LocalScanResult']) -> None:
 
         for local_scan_result in local_scan_results:
             for document_detections in local_scan_result.document_detections:
-                report_url = local_scan_result.report_url if local_scan_result.report_url else 'N/A'
                 for detection in document_detections.detections:
-                    table.set(REPORT_URL_COLUMN, report_url)
```
MarshalX, line 64: pls remove REPORT_URL_COLUMN constant too at the beginning of the file

MarshalX (approved):
lgtm

PR #208, opened 2024-02-05
CM-32048 - Fix scans runs for non-admin users

MarshalX:
drafter till BE fixes for secrets

### DIFF
--- a/cycode/cyclient/scan_client.py
+++ b/cycode/cyclient/scan_client.py
@@ -20,33 +20,32 @@ class ScanClient:
         self.scan_cycode_client = scan_cycode_client
         self.scan_config = scan_config
 
-        self._SCAN_CONTROLLER_PATH = 'api/v1/scan'
-        self._SCAN_CONTROLLER_PATH_SCA = 'api/v1/cli-scan'
+        self._SCAN_SERVICE_CONTROLLER_PATH = 'api/v1/scan'
+        self._SCAN_SERVICE_CLI_CONTROLLER_PATH = 'api/v1/cli-scan'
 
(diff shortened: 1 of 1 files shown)


artem-fedorov (approved):

artem-fedorov (approved):

cycode/cyclient/scan_client.py
```diff
@@ -20,33 +20,33 @@ def __init__(
...
 
-    def get_scan_controller_path(self, scan_type: str, should_use_scan_service: bool = False) -> str:
-        if should_use_scan_service:
-            return self._SCAN_CONTROLLER_PATH
-        if scan_type == consts.SCA_SCAN_TYPE:
-            return self._SCAN_CONTROLLER_PATH_SCA
+    def get_scan_controller_path(self, scan_type: str) -> str:
+        if scan_type == consts.INFRA_CONFIGURATION_SCAN_TYPE:
+            # we don't use async flow for IaC scan yet
+            return self._SCAN_SERVICE_CONTROLLER_PATH
```
MichalBor, line 36: @MarshalX so we never reach to here?

MarshalX: we reach it during my local tests. because iac team already implemented support for it. idk will it work with new CLI controller. but it worked with old controller. that's why I left it here

MichalBor: if we release version today, it will include the async iac scan?

MarshalX: today no. iac still goes to detector directly

MichalBor: cool, so its ok for now, but please make sure that when the async iac scan will be released it will use only the cli scan routes

MichalBor (approved):

PR #211, opened 2024-02-12
CM-29961 - Disable dependencies restore of SCA for sync scans

MarshalX:


artem-fedorov (approved):

MarshalX:
well, ig IDE plugins can pass 2 options eg `--sync --no-restore`

no-restore feature was added in https://github.com/cycodehq/cycode-cli/pull/154

PR #230, opened 2024-05-26
CM-30526 - Make Git executable optional

MarshalX:
idea: we don't need to work with git in all cases. for example, path scans could be executed without depending on git. git used to get commit history and so on.

how does it work: GitPython on import tries to find and verify git executable installed in the system. it raises an exception if it can't be found. we can catch this exception and provide a fake git provider, which will throw exceptions only on specific function calls instead of on the module import. all type hints from git module are hidden behind TYPE_CHECKING var

also, these exceptions go via our printers and depend on --output format now (json support for IDE plugins)

### DIFF
--- a/cycode/cli/utils/git_proxy.py
+++ b/cycode/cli/utils/git_proxy.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from functools import cache
+from functools import lru_cache
 from typing import TYPE_CHECKING, Optional, Type
 
 _GIT_ERROR_MESSAGE = """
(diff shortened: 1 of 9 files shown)


cycode/cli/utils/git_proxy.py
```diff
@@ -0,0 +1,82 @@
+from abc import ABC, abstractmethod
+from functools import lru_cache
+from typing import TYPE_CHECKING, Optional, Type
+
+_GIT_ERROR_MESSAGE = """
+Cycode CLI needs the git executable to be installed on the system.
+Git executable must be available in the PATH.
+Git 1.7.x or newer is required.
+You can help Cycode CLI to locate the Git executable
+by setting GIT_PYTHON_GIT_EXECUTABLE=<path/to/git> environment variable.
```
elsapet, line 10:
```suggestion
by setting the GIT_PYTHON_GIT_EXECUTABLE=<path/to/git> environment variable.
```

elsapet (approved):

elsapet (approved):

elsapet (approved):

PR #235, opened 2024-07-08
CM-37862 - Integrate Sentry

MarshalX:


MaorDavidzon (commented):
-

cycode/cli/utils/jwt_utils.py
```diff
@@ -0,0 +1,14 @@
+from typing import Tuple
+
+import jwt
+
+
+def get_user_and_tenant_ids_from_access_token(access_token: str) -> Tuple[str, str]:
+    payload = jwt.decode(access_token, options={'verify_signature': False})
```
elsapet, line 7: Curious why we don't verify the JWT signature here?

MarshalX: @elsapet because we can not expose our private key which is required to generate a signature to be able to compare it

elsapet (approved):

PR #246, opened 2024-08-19
CM-3937 - Extend tfplan scan result with module path

avishaiamiel:


doratias18 (approved):

cycode/cli/files_collector/iac/tf_content_generator.py
```diff
@@ -60,8 +68,10 @@ def _extract_resources(tfplan: str, filename: str) -> List[ResourceChange]:
         for resource_change in resource_changes:
             resources.append(
                 ResourceChange(
+                    module_address=resource_change.get('module_address'),
```
MarshalX, line 71: we are using `.get()` which will return `None` if the field does not exist. do we expect this? if so, should we mark these fields as `Optional[]` in `ResourceChange` model? Other fields that we had before do not used with `.get()`

cycode/cli/files_collector/iac/tf_content_generator.py
```diff
@@ -44,14 +43,23 @@ def _generate_tf_content(resource_changes: List[ResourceChange]) -> str:
...
     return resource_content
 
 
+def _get_resource_name(resource_change: ResourceChange) -> str:
+    parts = [resource_change.module_address, resource_change.name]
+
+    if resource_change.index is not None:
+        parts.append(str(resource_change.index))
+
+    return '-'.join(filter(None, parts)).replace('.', '-')
```
MarshalX, line 60: could you pls explain trick around `filter(None, someList`)?

avishaiamiel: I just wanted to concat only the parts that are not None

cycode/cli/files_collector/iac/tf_content_generator.py
avishaiamiel: Yes those are optional. Fixed.

cycode/cli/files_collector/iac/tf_content_generator.py
MarshalX: understood. IMO `[part for part in parts if part]` could be more readable because i had no clue that filter accepts `None` to the `fun` param. but as you want. maybe just add some comment around is enough

avishaiamiel: changed per your suggestion

MarshalX (approved):
good job!

PR #252, opened 2024-10-10
CM-40908 - Implement NuGet restore support for SCA

naftalicy:


cycode/cli/files_collector/sca/sca_code_scanner.py
```diff
@@ -104,7 +103,7 @@ def try_restore_dependencies(
             restore_dependencies_document.content = ''
         else:
             is_monitor_action = context.obj['monitor']
-            project_path = context.params['paths'][0]
+            project_path = context.params['path'][0]
```
MarshalX, line 106: Probably we want to handle both cases? In the case of `cycode scan repository` as I can see `path` will be right, but in the case of `cycode scan path` it will be `paths`. Pls check

naftalicy: Added checks for both

cycode/cli/files_collector/sca/sca_code_scanner.py
```diff
@@ -104,7 +103,11 @@ def try_restore_dependencies(
             restore_dependencies_document.content = ''
         else:
             is_monitor_action = context.obj['monitor']
-            project_path = context.params['paths'][0]
+
+            project_path = context.params.get('paths', [None])
+            if project_path is None:
+                project_path = context.params['path'][0]
```
MarshalX, line 109:
```suggestion
            project_path = context.params['path'] if 'path' in context.params else context.params['paths'][0]
```

cycode/cli/utils/path_utils.py
```diff
@@ -97,6 +97,10 @@ def change_filename_extension(filename: str, extension: str) -> str:
...
     if filename.startswith(os.sep):
         # remove leading slash to join the path correctly
-        filename = filename[len(os.sep) :]
+        filename = filename[len(os.sep):]
 
     return os.path.join(unique_id, filename)
+
+
+def get_path_from_context(context: dict) -> str:
+    return context.params['path'] if 'path' in context.params else context.params['paths'][0]
```
MarshalX, line 106: fixed context type; add make optional return type in case of missing path/paths

we need to add "import click" at the top

```suggestion
def get_path_from_context(context: click.Context) -> Optional[str]:
    path = context.params.get('path')
    if path is None and 'paths' in context.params:
        path = context.params['paths'][0]

    return path
```

naftalicy: Actually that was my first implementation at the beginning and that is why I wouldn't used "context.params['path'] if 'path' in context.params else context.params['paths'][0]"

I'll return it back
Check it out

MarshalX: I was so confused by your implementation because `project_path` is never will be `None` because of the default value set to array (`[None]`). and calling `[0]` from `path` and not from `paths`. But in the end, thank you so much for your patience and contribution!

MarshalX (approved):
great job!

PR #253, opened 2024-10-11
chore: support py3.13

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/193794


MarshalX:
hey @chenrui333, thank you so much for your effort. I understand how critical it is to homebrew, so I decided to not deal with ping-pong in this PR and did a proper project update here https://github.com/cycodehq/cycode-cli/pull/255. starting from the next release Python 3.13 will be supported and the patch could be removed from homebrew-core. thank you again!

PR #271, opened 2024-11-28
CM-42089 - Implement Ruby restore support for SCA

naftalicy:


cycode/cli/files_collector/sca/ruby/restore_ruby_dependencies.py
```diff
@@ -0,0 +1,30 @@
...
+from cycode.cli.files_collector.sca.base_restore_dependencies import BaseRestoreDependencies
+from cycode.cli.models import Document
+
+RUBY_PROJECT_FILE_EXTENSIONS = ['Gemfile']
+RUBY_LOCK_FILE_NAME = 'Gemfile.lock'
+
+
+class RestoreRubyDependencies(BaseRestoreDependencies):
+    def __init__(self, context: click.Context, is_git_diff: bool, command_timeout: int) -> None:
+        super().__init__(context, is_git_diff, command_timeout)
```
MarshalX, line 15: no need to override if the body is empty

cycode/cli/files_collector/sca/ruby/restore_ruby_dependencies.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def is_project(self, document: Document) -> bool:
+        return any(document.path.endswith(ext) for ext in RUBY_PROJECT_FILE_EXTENSIONS)
+
+    def get_commands(self, manifest_file_path: str) -> List[List[str]]:
+        return [['/opt/homebrew/opt/ruby/bin/bundle', '--quiet']]
```
MarshalX, line 21: this path will not work in many cases: ruby could be installed not via homebrew; homebrew does not exists on Windows

MarshalX (approved):

PR #276, opened 2025-01-27
CM-27209 - Add latest CLI version check

MarshalX:
DEMO:
image

Notes of implementation:
- It does not annoy users because it appears 1 time in a week for stable releases and 1 time in a day for pre-releases
- It always appears for `cycode version` and `cycode status` commands
- It does not appear for `--output json` to not break parsers
- It is possible to disable the version check using `cycode --no-update-notifier ...`
- It appears at the end of any stdout

### DIFF
--- a/cycode/cli/commands/version/version_checker.py
+++ b/cycode/cli/commands/version/version_checker.py
@@ -12,7 +12,7 @@ from cycode.cyclient.cycode_client_base import CycodeClientBase
 
 
 def _compare_versions(
-        current_parts: List[int],
+    current_parts: List[int],
     latest_parts: List[int],
     current_is_pre: bool,
     latest_is_pre: bool,
--- a/cycode/cli/commands/version/version_checker.py
+++ b/cycode/cli/commands/version/version_checker.py
@@ -1,0 +1,213 @@
+import os
+import re
+import time
+from pathlib import Path
+from typing import List, Optional, Tuple
+
+import click
+
+from cycode.cli.user_settings.configuration_manager import ConfigurationManager
+from cycode.cli.utils.path_utils import get_file_content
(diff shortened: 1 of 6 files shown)


cycode/cli/commands/version/version_checker.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def _update_last_check(self) -> None:
+        """Update the timestamp of the last update check.
+
+        Creates the cache directory if it doesn't exist and write the current timestamp
+        to the cache file. Silently handle any IO errors that might occur during the process.
+        """
+        try:
+            os.makedirs(os.path.dirname(self.cache_file), exist_ok=True)
+            with open(self.cache_file, 'w', encoding='UTF-8') as f:
```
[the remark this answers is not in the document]
MarshalX: #cycode_sast_ignore_here `self.cache_file` is our constant

cycode/cli/commands/version/version_checker.py
```diff
@@ -0,0 +1,213 @@
...
+        """Initialize the VersionChecker.
+
+        Sets up the version checker with PyPI API URL and configure the cache file location
+        using the global configuration directory.
+        """
+        super().__init__(self.PYPI_API_URL)
+
+        configuration_manager = ConfigurationManager()
+        config_dir = configuration_manager.global_config_file_manager.get_config_directory_path()
+        self.cache_file = Path(config_dir) / '.version_check'
```
gotbadger, line 75: Would this not end up in the project path? Could we write to the system temp dir instead?

MarshalX: it will be in `~/.cycode/` because of `global_config_file_manager`

PR #332, opened 2025-08-26
CM-52106 - Add .cycodeignore support

DmytroLynda:


cycode/cli/files_collector/walk_ignore.py
```diff
@@ -6,6 +6,7 @@
 
 _SUPPORTED_IGNORE_PATTERN_FILES = {  # oneday we will bring .cycodeignore or something like that
```
MarshalX: IG this comment is not needed anymore since we have 2 files already

DmytroLynda: Right. Fixed 👍

PR #352, opened 2025-10-08
CM-53944-Fix-docker-file-ignore-issue

galf16:


cycode/cli/files_collector/file_excluder.py
```diff
@@ -100,7 +104,8 @@ def _is_relevant_file_to_scan_common(self, scan_type: str, filename: str) -> boo
             )
             return False
 
-        if not self._is_file_extension_supported(scan_type, filename):
+        if (not self._is_file_extension_supported(scan_type, filename)
```
gotbadger: you can make this a lot more reabable by doing `if not (a or b)`

galf16: Fixed

PR #417, opened 2026-03-25
CM-61550: Handle HTTP 408 with a user-friendly slow connection message

mateusz-sterczewski:
Adds HttpRequestTimeoutError as a distinct subclass of RequestHttpError for 408 responses, so users see a specific message about slow connections instead of the generic "unable to complete scan" error.


elsapet (approved):
Nice 😎

PR #434, opened 2026-04-09
CM-62381-add-session-start-hook

RoniCycode:
- Change "ensure-auth" to "session-start" which includes conversation creation and MCP usage checks (once per session)


cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -0,0 +1,106 @@
...
+from cycode.cli.utils.get_api_client import get_ai_security_manager_client
+from cycode.logger import get_logger
+
+logger = get_logger('AI Guardrails')
+
+
+def _build_session_payload(payload: dict, ide: str) -> AIHookPayload:
+    """Build an AIHookPayload from a session-start stdin payload."""
+    if ide == AIIDEType.CLAUDE_CODE:
+        ide_version, model, _ = _extract_from_claude_transcript(payload.get('transcript_path'))
```
Ilanlido: we should probably get rid of _extract_from_claude_transcript now, i created it because other hooks didn't provide the model, as far as i know session start provides the model in stdin
Regarding ide_version, do we have it in stdin / claude.json file?

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -0,0 +1,106 @@
...
+        claude_config = load_claude_config()
+        ide_user_email = get_user_email(claude_config) if claude_config else None
+
+        return AIHookPayload(
+            event_name='session_start',
```
Ilanlido: no need - its not used later when creating conversation

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -0,0 +1,106 @@
...
+        logger.debug('Failed to initialize AI security client', exc_info=e)
+        return
+
+    # Step 4: Create conversation
+    try:
+        ai_client.create_conversation(session_payload)
+    except Exception as e:
+        logger.debug('Failed to create conversation during session start', exc_info=e)
+
+    # Step 5: Report data flow (MCP servers, Claude Code only)
```
Ilanlido: we should handle cursor too, also extract to func

cycode/cyclient/ai_security_manager_client.py
```diff
@@ -17,6 +17,7 @@ class AISecurityManagerClient:
 
     _CONVERSATIONS_PATH = 'v4/ai-security/interactions/conversations'
     _EVENTS_PATH = 'v4/ai-security/interactions/events'
+    _DATA_FLOW_PATH = 'v4/ai-security/interactions/data-flow'
```
Ilanlido: i dont know if i like that route name

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -0,0 +1,106 @@
...
+            handle_auth_exception(ctx, err)
+            return
+    else:
+        logger.debug('Already authenticated')
+
+    # Step 2: Read stdin payload (backward compat: old hooks pipe no stdin)
+    if sys.stdin.isatty():
+        logger.debug('No stdin payload (TTY), skipping session initialization')
+        return
+
```
Ilanlido, line 127: can you explain

cycode/cyclient/ai_security_manager_client.py
RoniCycode: 🤷

cycode/cli/apps/ai_guardrails/session_start_command.py
RoniCycode: What it does: sys.stdin.isatty() returns True when stdin is attached to a terminal — i.e. nothing is piped in. If we didn't have this guard, sys.stdin.read() on the next line would block forever waiting   
  for input that never arrives.                                                                                                                                                                                
                                                                                                                                                                                                               
  Why it's needed:                                                                                                                                                                                             
                                                                                                                                                                                                               
  1. Manual/interactive invocation. If you run cycode ai-guardrails session-start in a shell to debug it, stdin is the terminal. Without this check, the command would hang.                                   
  2. Backward compat with the old ensure-auth hook. In __init__.py we kept ensure-auth as a deprecated alias:
  app.command(hidden=True, name='ensure-auth', ...)(session_start_command)                                                                                                                                     
  2. Users who installed hooks before this branch have cycode ai-guardrails ensure-auth wired into their IDE config. The old command only ran auth and didn't read stdin. After they upgrade the CLI, that same
   hook now invokes session_start_command. The guard ensures that even if the invocation context somehow doesn't provide a JSON payload on stdin, we still do the auth step (which is what the old command did)
   and then gracefully skip the conversation/MCP reporting steps instead of hanging.                                                                                                                           
                                                                                                                                                                                                               Flow:                                                                                                                                                                                                        
  - Hook invoked by IDE → stdin is a pipe → isatty() is False → read JSON, create conversation, report MCP
  - Hook invoked with no stdin (manual / edge case) → isatty() is True → auth happens above, then we return early                                                                                              
                                                                                                                 
  If you want to make the intent clearer, the comment could be:                                                                                                                                                
  - No piped input: would block forever on read(). Happens when run manually,                                                                                                                                  
  -  or with the legacy `ensure-auth` hook which didn't supply stdin.

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -19,22 +20,19 @@
 def _build_session_payload(payload: dict, ide: str) -> AIHookPayload:
     """Build an AIHookPayload from a session-start stdin payload."""
     if ide == AIIDEType.CLAUDE_CODE:
-        ide_version, model, _ = _extract_from_claude_transcript(payload.get('transcript_path'))
```
Ilanlido: remove this function definition and conversation creation on every hook

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -19,22 +20,19 @@
...
 
         return AIHookPayload(
-            event_name='session_start',
             conversation_id=payload.get('session_id'),
             ide_user_email=ide_user_email,
-            model=payload.get('model') or model,
+            model=payload.get('model'),
             ide_provider=AIIDEType.CLAUDE_CODE.value,
-            ide_version=ide_version,
+            ide_version=None,
```
Ilanlido: we still need to get the version

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -43,6 +41,28 @@ def _build_session_payload(payload: dict, ide: str) -> AIHookPayload:
...
+    return get_mcp_servers(config) or {} if config else {}
+
+
+def _report_data_flow(ai_client, ide: str) -> None:
+    """Report IDE MCP servers to the AI security manager. Never raises."""
+    mcp_servers = _get_mcp_servers_for_ide(ide)
+    if not mcp_servers:
+        return
+    try:
+        ai_client.report_data_flow(mcp_servers)
```
Ilanlido: rename

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -0,0 +1,118 @@
...
+    except Exception as e:
```
Ilanlido, line 94: you already have try catch inside report_data_flow

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -22,13 +22,14 @@ def _build_session_payload(payload: dict, ide: str) -> AIHookPayload:
     if ide == AIIDEType.CLAUDE_CODE:
         claude_config = load_claude_config()
         ide_user_email = get_user_email(claude_config) if claude_config else None
+        ide_version, _, _ = _extract_from_claude_transcript(payload.get('transcript_path'))
```
Ilanlido: do we have a better way to get claude version? i want to get rid of this function and also stop creating conversations for each hook

tests/cli/commands/ai_guardrails/test_session_start_command.py
```diff
@@ -218,29 +220,36 @@ def test_claude_code_reports_mcp_servers(
         'filesystem': {'command': 'npx', 'args': ['-y', '@modelcontextprotocol/server-filesystem']},
     }
     mock_load_config.return_value = {'oauthAccount': {'emailAddress': 'u@e.com'}, 'mcpServers': mcp_servers}
+    enabled_plugins = {'cycode-dev@cycode-marketplace': True}
```
Ilanlido: what does this json contains? does it contain the mcps as well?

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -0,0 +1,144 @@
...
+    else:
+        enriched_plugins = {}
+
+    return mcp_servers, enriched_plugins
+
+
+def _get_cursor_session_context() -> tuple[dict, dict]:
+    """Return (mcp_servers, enabled_plugins) for Cursor. Cursor has no plugin system."""
+    config = load_cursor_config()
+    mcp_servers = get_mcp_servers(config) or {} if config else {}
```
elsapet: Suggestion to use parens here to make it clearer and less ambiguous. Also, maybe we want to ensure a dict as we do on line 59?

```suggestion
    mcp_servers = dict(get_mcp_servers(config) or {}) if config else {}
```

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -0,0 +1,144 @@
...
+    return mcp_servers, {}
+
+
+def _report_session_context(ai_client, ide: str) -> None:
+    """Report IDE session context to the AI security manager. Never raises."""
```
elsapet, line 82: > Never raises.

This comment confuses me because there's nothing in the `_report_session_context` method itself that guarantees this. 

If we want to ensure this behaviour here, I would wrap this method logic itself in a `try` / `except` block. What do you think?

cycode/cli/apps/ai_guardrails/scan/claude_config.py
```diff
@@ -42,3 +43,121 @@ def get_user_email(config: dict) -> Optional[str]:
...
+        return None
+    try:
+        content = path.read_text(encoding='utf-8')
+        return json.loads(content)
+    except Exception as e:
+        logger.debug('Failed to load Claude settings file', exc_info=e)
+        return None
+
+
+def get_enabled_plugins(settings: dict) -> Optional[dict]:
```
elsapet: Where is this method used?

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -0,0 +1,144 @@
...
+from cycode.cli.apps.ai_guardrails.consts import AIIDEType
+from cycode.cli.apps.ai_guardrails.scan.claude_config import (
+    get_mcp_servers,
+    get_user_email,
+    load_claude_config,
+    load_claude_settings,
+    resolve_plugins,
+)
+from cycode.cli.apps.ai_guardrails.scan.cursor_config import load_cursor_config
+from cycode.cli.apps.ai_guardrails.scan.payload import AIHookPayload, _extract_from_claude_transcript
```
elsapet: If we're importing `_extract_from_claude_transcript` here, should we make it a public function?

cycode/cli/apps/ai_guardrails/__init__.py
```diff
@@ -18,6 +18,9 @@
     name='scan',
     short_help='Scan content from AI IDE hooks for secrets (reads JSON from stdin).',
 )(scan_command)
-app.command(hidden=True, name='ensure-auth', short_help='Ensure authentication, triggering auth if needed.')(
-    ensure_auth_command
+app.command(hidden=True, name='session-start', short_help='Handle session start: auth, conversation, session context.')(
+    session_start_command
+)
+app.command(hidden=True, name='ensure-auth', short_help='[Deprecated] Alias for session-start.')(
+    session_start_command
```
elsapet: I assume because these are hidden commands, there are no related documentation updates required 👍 

Also, for interest, is it planned to clean up / remove the deprecated command in a later PR?

elsapet (commented):
Some comments for your consideration but no major blockers IMO 🙏

cycode/cli/apps/ai_guardrails/__init__.py
RoniCycode: Yes, and yes :)

cycode/cli/apps/ai_guardrails/session_start_command.py
```diff
@@ -0,0 +1,118 @@
...
+
+    try:
+        ai_client = get_ai_security_manager_client(ctx)
+    except Exception as e:
+        logger.debug('Failed to initialize AI security client', exc_info=e)
+        return
+
+    # Step 4: Create conversation
+    try:
+        ai_client.create_conversation(session_payload)
```
Ilanlido, line 145: now that i think about it, if we call create conversation here, we should do it on every new session hook ( currently we filter for only `startup` https://github.com/cycodehq/cycode-cli/blob/main/cycode/cli/apps/ai_guardrails/consts.py#L121

cycode/cli/apps/ai_guardrails/session_start_command.py
RoniCycode: no, there isn't a better source. Checked the Claude Code hook docs — no env var, no field in the SessionStart payload. Only alternatives are (a) shell out to claude       
  --version (worse than transcript read), (b) drop ide_version from session-start entirely and let it get filled in later by scan events (which already read the transcript for model/generation_id).

elsapet (approved):

PR #435, opened 2026-04-12
CM-62578: Expose Cycode API v4 through CLI commands

MaorDavidzon:
## Summary
- Expose Cycode API v4 as CLI commands, dynamically generated from the OpenAPI spec
- OpenAPI spec fetched from `app.cycode.com` and cached locally for 24h
- All new commands marked as `[EXPERIMENT]`
- Uses the CLI's standard auth client (same as scan/report commands)

### Projects
```bash
cycode projects list --page-size 10 --query "Frontend"
cycode projects view 29303
cycode projects collisions --project-id 123
cycode projects collisions-count --project-id 123
cycode projects assets --project-id 123
```

### Violations
```bash
cycode violations list --severity Critical --status Open --page-size 5
cycode violations list --category SAST --sort-by "-risk_score"
cycode violations view <violationId>
cycode violations count --severity High --category SecretDetection
cycode violations scan <scanId>
```

### Workflows
```bash
cycode workflows list --query "classification"
cycode workflows view <workflowId>
cycode workflows count
cycode workflows jobs <workflowId>
cycode workflows jobs-count <workflowId>
cycode workflow-jobs list
cycode workflow-jobs view <jobId>
```

### Members
```bash
cycode members list --page-size 5
cycode members view <memberId>
cycode members roles
cycode members me
```

### Compliance
```bash
cycode frameworks list
cycode frameworks view <frameworkId>
cycode controls list --framework-id 1
cycode controls view <controlId>
cycode stats framework <frameworkId>
cycode stats posture
```

### Audit logs
```bash
cycode audit-logs list --from 2024-01-01 --to 2024-12-31 --action Modified
cycode audit-logs view <id>
```

### Scans & statistics
```bash
cycode scans list --page-size 5
cycode scans view <scanId>
cycode scans statistics --from 2024-01-01
cycode cli-scan-statistics scan-count --from 2024-01-01
cycode cli-scan-statistics top-cli-users --from 2024-01-01
```

### Other resources
```bash
cycode labels list
cycode brokers list
cycode tokens tenant
cycode reports list
cycode secret-value-exclusions exclusions
```

### Discovery
```bash
cycode --help                        # all 46 command groups
cycode violations --help             # subcommands for violations
cycode violations list --help        # all filter flags for listing violations
```

## Architecture
- OpenAPI spec fetched from `app.cycode.com/v4/api-docs/cycode-api-swagger.json` with a 3s timeout
- If fetch times out, a background thread continues so the cache is warm for the next run
- Cached runs have zero network overhead
- Command names derived from URL path structure (not English summary parsing)
  - `list` for collections, `view` for single resources
  - Sub-paths become command names: `/violations/count` → `count`
- Path parameters are URL-encoded to prevent path traversal
- Auth uses `CycodeTokenBasedClient` (same as all other CLI commands)

## Test plan
- [x] `cycode projects list --page-size 2` returns project data
- [x] `cycode violations list --severity Critical --status Open` returns filtered violations
- [x] `cycode projects view <ID>` returns single project (path param)
- [x] `cycode violations count --severity High` returns count
- [x] `cycode workflows list` returns workflows
- [x] `cycode members roles` returns roles
- [x] `cycode --help` shows API command groups with [EXPERIMENT] label
- [x] All 645 existing tests pass
- [x] Graceful fallback when not authenticated (no API commands, no crash)
- [x] Spec caching works (second run has no fetch)
- [x] 3s timeout prevents blocking on slow networks

Jira: [CM-62578](https://cycode.atlassian.net/browse/CM-62578)

[CM-62578]: https://cycode.atlassian.net/browse/CM-62578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gotbadger (changes requested):
I think we should put all the commands under an `api` namespace so like `cycode api violations` right now its really cluttering up the main commands and means we can have naming collisions. This is similar to what github is doing for their CLI. Also if we do this it means when we run unrelated tasks we wont have pull the swagger file, we should try and do this as when CLI is run in CI/CD it will never be hitting cache.


Here is an example of where you get some annoying logging in a command due to the call for openAPI being made in the background.
```
poetry run cycode status                                                                                                       
[04/13/26 10:34:02] WARNING  [API Command] Could not load OpenAPI spec: Cycode credentials not found. Run `cycode auth` first, or set CYCODE_CLIENT_ID and CYCODE_CLIENT_SECRET environment variables.              api_command.py:128
Program: cycode
Version: 0.0.0
Os: Darwin
Arch: arm64
Python version: 3.13.7
Installation id: xxxx
App url: https://app.cycode.com
Api url: https://api.cycode.com
Is authenticated: False
User id: None
Tenant id: None
Supported modules:
  Secret scanning: False
  Sca scanning: False
  Iac scanning: False
  Sast scanning: False
  Ai large language model: False
....
```

Lastly we will need to add documentation to the Readme for these new commands. Having it under the api namespace will make this a bit easier I think in explaining generally what it does since its dynamic

MaorDavidzon:
@gotbadger Great feedback, thank you — all three points are valid.

**You're right about the concrete issues:**
1. Top-level registration causes namespace collisions with curated commands (`scan`, `auth`, `status`, `report`)
2. Spec is loaded on every invocation, including unrelated commands like `cycode status` — bad in CI where the cache is always cold
3. The unauthenticated warning leaking into `cycode status` output is a real bug

**On namespacing, going with `platform` instead of `api`:**

I looked at how other CLIs handle this. `gh api` and `glab api` are *raw URL passthroughs* (`gh api repos/owner/repo/issues`) — what we built is structurally different: typed commands with proper args and `--option` flags generated from the spec. The closer analogs are `aws <service>`, `gcloud <product>`, `az <group>`, `oci <service>` — all of which put structured generated commands at the top level (because they bundle the spec).

Since we can't easily bundle the spec (it's tenant-aware and changes more frequently than CLI releases), namespacing is the right call. `api` telegraphs "raw escape hatch" the way `gh api` does, which doesn't match what these commands actually are. **`platform`** frames it as "the Cycode platform's full surface," reads naturally (`cycode platform projects list`), doesn't leak implementation detail (OpenAPI/spec/swagger), and ages better if we ever add POST/PUT.

**Plan:**
- Move all generated commands under `cycode platform <resource> <action>`
- Lazy-load the spec only when the user enters the `platform` group — no fetch on `cycode scan`, `cycode status`, etc.
- Drop the startup warning entirely (only surface errors when the user is actually inside `platform`)
- Add a README section explaining the `platform` namespace and how command discovery works

cycode/cli/apps/api/openapi_spec.py
```diff
@@ -0,0 +1,176 @@
...
+    Uses CycodeTokenBasedClient for auth and retries. The spec is served from the app URL,
+    so we create a client with app_url as base instead of the default api_url.
+    """
+    from cycode.cyclient.cycode_token_based_client import CycodeTokenBasedClient
+
+    cid, csecret = resolve_credentials(client_id, client_secret)
+
+    # Create a token-based client pointed at app URL (spec is served from app, not api)
+    client = CycodeTokenBasedClient(cid, csecret)
+    client.api_url = cyclient_config.cycode_app_url
```
gotbadger, line 99: **Bug: `api_url` override breaks token refresh** `client.api_url = cyclient_config.cycode_app_url` is set before any request is made. When the bearer token
  is expired, the client tries to refresh it by POSTing to `self.api_url/api/v1/auth/api-token` — but that now points at `app.cycode.com` instead of `api.cycode.com`, so auth fails and
  the spec fetch fails with a misleading error. It works on a subsequent retry if another command (e.g. `cycode status`) ran in between, because that command refreshes the token via a
  correctly-configured client and caches it to disk. The next `cycode platform` call loads the fresh token and never needs to refresh, so the `api_url` override only affects the spec
  GET (which correctly targets the app host). Fix: ensure the token is refreshed before overriding `api_url`, or avoid mutating it entirely.

cycode/cli/apps/api/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+
+class PlatformGroup(click.Group):
+    """Lazy-loading Click group for `cycode platform` subcommands.
+
+    The OpenAPI spec is only fetched when the user actually invokes
+    `cycode platform ...` (or asks for its help). Unrelated commands like
+    `cycode scan` or `cycode status` never trigger a spec fetch.
+    """
+
+    _loaded: bool = False
```
gotbadger: This is a class variable not an instance variable?

cycode/cli/apps/api/openapi_spec.py
```diff
@@ -0,0 +1,176 @@
...
+
+    spec_path = _OPENAPI_SPEC_PATH.lstrip('/')
+    logger.info('Fetching OpenAPI spec from %s/%s', cyclient_config.cycode_app_url, spec_path)
+
+    try:
+        response = client.get(spec_path)
+        spec = response.json()
```
gotbadger, line 106: Suggestion: might be worth validating the file in some way to make sure it actually looks like an openapi spec not an error message or something returned unexpectidly, don't think its critical however.

gotbadger (commented):
Think we are mostly good structure wise here now, couple of things I spotted.

cycode/cli/apps/api/openapi_spec.py
MaorDavidzon: Great catch — this was a real bug. When the bearer token is expired, `_request_new_access_token()` POSTs to `self.api_url/api/v1/auth/api-token`, which after the override pointed at `app.cycode.com` instead of `api.cycode.com`.

Fixed in e604a38 by calling `client.get_access_token()` *before* overriding `api_url`. This ensures the token refresh targets `api.cycode.com` (correct), and only the subsequent spec GET goes to `app.cycode.com`.

cycode/cli/apps/api/__init__.py
MaorDavidzon: Good catch — it was indeed a class variable, so all instances would share the same `_loaded` flag. Moved it to `__init__` as an instance variable in 6e3cd86.

cycode/cli/apps/api/openapi_spec.py
MaorDavidzon: Good idea — added a basic validation in 9044f66. The spec is now rejected (and not cached) if the response is not a dict or is missing the `"paths"` key. This prevents caching error pages or unexpected HTML responses for 24 hours.

gotbadger (approved):

PR #452, opened 2026-05-06
CM-63882 - Added scanType validation

aaron-butler-cy-int:
the scan command use to accept multiple -t flags but would only run the last scan listed in the command. Added validation that only one -t flag was in the command.


cycode/cli/apps/scan/scan_command.py
```diff
@@ -126,6 +126,16 @@ def scan_command(
...
     """
+    if len(scan_type) > 1:
+        raise typer.BadParameter(
+            f'Only one scan type can be specified per command. '
+            f'Got: {", ".join(str(t) for t in scan_type)}. '
+            f'Run a separate command for each scan type.',
+            param_hint='-t/--scan-type',
+        )
+
+    resolved_scan_type = scan_type[0]
```
omer-roth: Try using a parameter callback to perform the validation, we can then reuse it later:
https://typer.tiangolo.com/tutorial/options/callback-and-context/

tests/cli/commands/scan/test_scan_command.py
```diff
@@ -25,6 +27,20 @@ def _invoke_result_callback(ctx: click.Context) -> int:
     return exc_info.value.exit_code
 
 
+class TestScanCommand:
+    def test_multiple_scan_types_rejected(self) -> None:
+        result = CliRunner().invoke(app, ['scan', '-t', 'iac', '-t', 'sast', 'path', '.'])
+        assert result.exit_code == 2
```
omer-roth: exit code 2 is used only if `--stop-on-error` flag is provided

```suggestion
        assert result.exit_code == 1
```

omer-roth (changes requested):

omer-roth (approved):

PR #475, opened 2026-06-22
Add supply-chain cooldown for package updates

darren-dooley:
## Why

Freshly published package versions are the highest-risk window for supply-chain attacks (compromised releases are often yanked within days of publication). Adding a cooldown means new releases must age before they can be adopted, giving the ecosystem time to catch and pull bad versions.

## What

- **`poetry.toml`** (new): sets `solver.min-release-age = 7`, so Poetry's solver ignores any release younger than 7 days when resolving the lock. This is project-local config, applied for everyone resolving in this repo. Requires Poetry >= 2.4.0 (the version the key was introduced in).
- **`pyproject.toml`**: bumped `requires-poetry` from `>=2.0` to `>=2.4.0` so the cooldown key is guaranteed to be honored.
- **`.github/dependabot.yml`**: added `cooldown.default-days: 10` to both the `github-actions` and `pip` ecosystems. Dependabot's window is intentionally longer than Poetry's 7-day floor so any version Dependabot proposes is already old enough for the solver to accept, avoiding PRs the lockfile resolution would reject as too young.

## Test case ran locally:

```bash
# Run on 22 Jun 2026
uvx poetry@2.4.0 add certifi  # resolves + installs under the repo's min-release-age = 7
uvx poetry@2.4.0 run python -c "import certifi; print(certifi.__version__)"  # -> 2026.05.20 (aged), not today's latest
uvx poetry@2.4.0 show --latest | grep '^certifi'  # latest with cooldown ON -> 2026.5.20
POETRY_SOLVER_MIN_RELEASE_AGE=0 uvx poetry@2.4.0 show --latest | grep '^certifi'  # latest with cooldown OFF -> 2026.6.17
uvx poetry@2.4.0 remove certifi  # clean up (demo only, not a real dependency)
```


pyproject.toml
```diff
@@ -29,7 +29,7 @@ cycode = "cycode.cli.app:app"
 repository = "https://github.com/cycodehq/cycode-cli"
 
 [tool.poetry]
-requires-poetry = ">=2.0"
+requires-poetry = ">=2.4.0"
```
omer-roth: In version 2.4+ Poetry dropped support for Python 3.9.

While Python 3.9 reached EOL, it is still shipped as default for most operating systems and so we need to maintain backward compatibility

darren-dooley: @omerr-cycode Good catch, thanks. I've reverted requires-poetry back to >=2.0 so the Python 3.9 toolchain keeps working, and left min-release-age in poetry.toml for devs using a modern version of Poetry.

The cooldown will still be applied on the Dependabot side for all versions of Poetry.

omer-roth (approved):

omer-roth:
One last thing @darren-dooley - please [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and I'll merge the request.

Thank you for updating this one

omer-roth:
Changed merge via PR https://github.com/cycodehq/cycode-cli/pull/483.
Closing this one

darren-dooley:
Hi @omer-roth, thanks for getting that one merged!

PR #478, opened 2026-06-28
CM-67195: Add Bun package manager support to SCA local scans

AradTraub:
## What & why

Enables `cycode scan -t sca` to support **Bun** projects during local SCA scans, so CLI users get the same coverage as platform scans. Closes the CLI portion of Bun Phase 1 (CM-67195).

Bun is now a first-class package manager alongside npm, Yarn, and Pnpm — the implementation mirrors the existing Pnpm/Yarn restore handlers.

## Changes

- **New `RestoreBunDependencies` handler** — detects a Bun project via `bun.lock` or a `packageManager` / `engines` bun signal in `package.json`. If `bun.lock` already exists it's read directly; otherwise `bun install --ignore-scripts` is run to generate it (then cleaned up).
- **Bun ≥ 1.2 enforcement** — only Bun 1.2+ emits the text-based `bun.lock` we parse (older Bun writes binary `bun.lockb`). Before generating a lockfile the handler verifies `bun --version`; on older/missing Bun it logs a warning and skips the restore rather than producing an unparseable lockfile. An already-present `bun.lock` skips the check (its existence implies Bun ≥ 1.2).
- **npm fallback guard** — `bun.lock` added to npm's alternative-lockfile guard so npm doesn't claim Bun projects, and the Bun handler is registered ahead of the npm fallback.
- **`consts.py`** — `bun.lock` added to the SCA supported files and the npm ecosystem map.
- **Tests** — unit coverage for detection, version gating (supported / old / missing Bun), direct-read of existing lockfile, and generated-lockfile cleanup; plus an "npm skips Bun project" test.

## Acceptance criteria

- ✅ Bun project detected via `bun.lock` → `bun install` restores deps before scanning
- ✅ Bun recognized alongside npm / Yarn / Pnpm
- ✅ CLI does not fall back to npm restore for Bun projects (when a lockfile is present)
- ✅ Unit tests cover the Bun restore flow

## Testing

- Full SCA test suite passes (173 tests, incl. 21 new for Bun); `ruff check` and `ruff format` clean.
- Verified end-to-end against a live environment: existing-`bun.lock` project reads the lockfile directly (npm skipped); no-lockfile project runs the version gate + `bun install --ignore-scripts` and generates/parses `bun.lock`.

## Note

There is a documented NOTE in `restore_npm_dependencies.py::is_project`: the npm guard is lockfile-on-disk based, so a project declaring `packageManager: "bun@..."` with *no* lockfile is claimed by both handlers (pre-existing behavior shared by pnpm/yarn). Left as-is for consistency; tracked on CM-67195.


cycode/cli/files_collector/sca/base_restore_dependencies.py
```diff
@@ -157,6 +175,52 @@ def get_any_restore_file_already_exist(self, document: Document, restore_file_pa
...
+            return False
+        if version < minimum_version:
+            logger.warning(
+                'Unsupported %s version; %s+ is required to restore dependencies, %s',
+                tool_name,
+                minimum_str,
+                {'detected_version': '.'.join(str(part) for part in version)},
+            )
+            return False
+        return True
```
omer-roth: Is this function relevant for anything other than Bun?

If not, I'd move to logic to `_indicates_bun`

tests/cli/files_collector/sca/npm/test_restore_bun_dependencies.py
```diff
@@ -0,0 +1,189 @@
...
+
+from cycode.cli.files_collector.sca.npm.restore_bun_dependencies import (
+    BUN_LOCK_FILE_NAME,
+    MINIMUM_BUN_VERSION,
+    RestoreBunDependencies,
+)
+from cycode.cli.models import Document
+
+# The minimum-version gate lives in the base class; patch shell/execute_commands there.
+_BASE_MODULE = 'cycode.cli.files_collector.sca.base_restore_dependencies'
```
omer-roth: Assuming we're moving the minimum-version gate to bun, is this still relevant?

cycode/cli/files_collector/sca/base_restore_dependencies.py
AradTraub: Good call — it's only used by Bun. Moved the version check back into `restore_bun_dependencies.py` (`_parse_bun_version` + `_is_supported_bun_version`) and reverted `BaseRestoreDependencies` to its original form. (f4ad542)

tests/cli/files_collector/sca/npm/test_restore_bun_dependencies.py
AradTraub: Right — reverted. The gate now lives in the Bun module again, so the version-parse/gate tests moved back to `test_restore_bun_dependencies.py` and patch the Bun module's `shell`; the base test module is unchanged from main. (f4ad542)

cycode/cli/files_collector/sca/npm/restore_npm_dependencies.py
```diff
@@ -23,6 +23,15 @@ def is_project(self, document: Document) -> bool:
...
         this one in the handler list. This handler is the npm fallback.
+
+        NOTE: this guard only excludes a project when an alternative lockfile is *physically
+        present on disk*. It does not inspect the `packageManager`/`engines` signal in
+        package.json. So a project that declares e.g. `packageManager: "bun@..."` (or pnpm)
+        but has no lockfile yet is claimed by BOTH the dedicated handler and this npm fallback,
+        and both restores run. This is pre-existing behavior shared by pnpm/yarn/bun and is
+        accepted for now (a real Bun/pnpm project ships a lockfile, so npm correctly skips).
+        If this ever needs tightening, also skip here when package.json declares a non-npm
+        packageManager/engines signal. Tracked on CM-67195.
```
omer-roth:
```suggestion
        packageManager/engines signal. 
```

omer-roth (dismissed):

omer-roth (approved):

PR #493, opened 2026-07-12
CM-68342: sweep all IDEs' session context and dedup reports

Ilanlido:
Session start now reports the device context unconditionally (previously a machine with no MCPs in the triggering IDE never created a device entity) and collects MCP configs from every registered IDE, not just the triggering one, sent as the config_files list. Unchanged payloads are skipped via a sha256 cache (~/.cycode/.session-context-cache, tenant-aware, 7-day TTL, written only after a successful send) so the common session start makes one API call instead of two.

Also: resolve plugins of git/github-sourced marketplaces through ~/.claude/plugins/cache (previously only directory-type marketplaces resolved, so their MCPs were silently missing), normalize plugin MCP content to the canonical {"mcpServers": ...} shape, and drop the legacy global_config_file field (the backend keeps its fallback for older CLIs).

### DIFF
--- a/cycode/cli/apps/ai_guardrails/ides/__init__.py
+++ b/cycode/cli/apps/ai_guardrails/ides/__init__.py
@@ -34,6 +34,25 @@ def get_ide(name: str) -> IDE:
     return ide
 
 
+def collect_all_session_contexts() -> tuple[dict[str, dict], dict]:
+    """Sweep every registered IDE's session context, regardless of which IDE triggered the hook.
+
+    Returns ``(config_files_by_ide, plugins)``: the global MCP config file of each IDE that has
+    one (keyed by IDE name), and the enabled plugins merged across IDEs (first registered IDE
+    wins on a duplicate plugin key - plugins are IDE-agnostic marketplace artifacts).
+    """
+    config_files_by_ide: dict[str, dict] = {}
+    plugins: dict = {}
+    for ide in IDES.values():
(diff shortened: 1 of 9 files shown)


omer-roth:
1. Cache format & compatibility: Cache stored at ~/.cycode/.session-context-cache with tenant-aware TTL. If format changes, old cache entries become harmless misses (no corruption risk).

2. Tenant ID dependency: Cache keyed by tenant_id from auth_info. If tenant ID is None/missing, caching still works but keys all nulls together. Tests verify this.

3. Plugin marketplace assumption: Assumes git/github plugins are cached at ~/.claude/plugins/cache/ (or ~/.codex/plugins/cache/). If Claude/Codex changes cache layout, plugins silently become unreachable (non-breaking, but incomplete).

4. Deterministic payload ordering: digest uses json.dumps(sort_keys=True) but config_files is constructed from dict.values() (insertion order of IDES). If IDES registration order changes (or if different Python versions/packaging cause different ordering), identical inventories could produce different digests and cause unnecessary resends or missed skips. Make the ordering explicit/deterministic before hashing (e.g., sort config_files by 'path' or include IDE name in each entry and sort by that).

5. Determinism in resolve_cached_plugin_dir: max(candidates, key=lambda d: d.stat().st_mtime) uses mtime. That is fine but could be brittle if two entries have same mtime. Acceptable, but consider stable tie-breaker (e.g., max by (mtime, name)).

omer-roth (approved):

PR #504, opened 2026-07-27
CM-68943 Send ai-guardrails hook context with the scan so report-mode findings become violations

Ilanlido:
The scan now carries the hook's context inside the generic scan_parameters metadata bag under an `ai_guardrails` key, which lets secret-detector attribute the finding to the developer's device and, in report mode, promote it to a real violation instead of a scan-scoped detection.

- `GuardrailsMode` replaces `InstallMode`: the same report/block pair describes both the installed mode and a single event's effective mode, which `get_effective_mode` now derives from the policy in one call.
- A generation id is minted per hook event when the IDE dialect does not supply one (Copilot), since it scopes the violation's identity.
- The device serial is cached in a temp file, keyed by username, so repeated hook invocations don't re-run the platform lookup.
- Copilot's sessionStart hook drops its timeoutSec.


cycode/cli/utils/host_info.py
```diff
@@ -104,6 +119,27 @@ def get_serial_number() -> Optional[str]:
...
+        return _serial_number_cache_path().read_text(encoding='utf-8').strip() or None
+    except Exception:
+        return None
+
+
+def _write_serial_number_cache(serial: str) -> None:
+    try:
+        cache_path = _serial_number_cache_path()
+        cache_path.parent.mkdir(parents=True, exist_ok=True)
+        cache_path.write_text(serial, encoding='utf-8')
```
omer-roth: Possible file corruption since Path.write_text is not atomic. 
Recommend writing to a temp file and doing `os.replace` to avoid corruption

cycode/cli/utils/host_info.py
```diff
@@ -104,6 +119,27 @@ def get_serial_number() -> Optional[str]:
...
+    try:
+        return _serial_number_cache_path().read_text(encoding='utf-8').strip() or None
+    except Exception:
+        return None
+
+
+def _write_serial_number_cache(serial: str) -> None:
+    try:
+        cache_path = _serial_number_cache_path()
+        cache_path.parent.mkdir(parents=True, exist_ok=True)
```
omer-roth: When creating a cache in the FS, please ensure only user running the command has permissions (600) to avoid leaking data between users of the same machine

cycode/cli/utils/host_info.py
```diff
@@ -104,6 +119,27 @@ def get_serial_number() -> Optional[str]:
...
+
+
+def _read_serial_number_cache() -> Optional[str]:
+    try:
+        return _serial_number_cache_path().read_text(encoding='utf-8').strip() or None
+    except Exception:
+        return None
+
+
+def _write_serial_number_cache(serial: str) -> None:
```
omer-roth, line 137: Please ensure we're not writing PII into this cache (serial numbers can be considered PII)

cycode/cli/apps/ai_guardrails/scan/handlers.py
```diff
@@ -295,6 +311,36 @@ def get_handler_for_event(event_type: str) -> Optional[HandlerFn]:
...
+    scan_parameters = get_scan_parameters(ctx, paths)
+    scan_parameters.setdefault('metadata', {})['ai_guardrails'] = {
+        'mode': effective_mode.value,
+        'ide_provider': payload.ide_provider,
+        'detection_source': SECRETS_BLOCK_REASON_BY_EVENT_TYPE[event_type].value,
+        'device_id': get_serial_number(),
+        'device_hostname': get_hostname(),
+        'conversation_id': payload.conversation_id,
+        'generation_id': payload.generation_id,
+        'ide_user_email': payload.ide_user_email,
```
omer-roth, line 337: `device_id`, `device_hostname` and `ide_user_email` can be considered PII, acknowledge this is the intention

cycode/cli/apps/ai_guardrails/ides/copilot.py
```diff
@@ -333,9 +333,7 @@ def entry(command: str) -> dict:
         return {
             'version': 1,
             'hooks': {
-                'sessionStart': [
-                    {'type': 'command', 'command': _SESSION_START_COMMAND, 'timeoutSec': _HOOK_TIMEOUT_SEC}
-                ],
+                'sessionStart': [{'type': 'command', 'command': _SESSION_START_COMMAND}],
```
omer-roth, line 336: timeout was removed, is this intentional?

cycode/cli/apps/ai_guardrails/scan/handlers.py
```diff
@@ -295,6 +311,36 @@ def get_handler_for_event(event_type: str) -> Optional[HandlerFn]:
...
 
 
+def get_effective_mode(policy: dict, feature_config: dict) -> GuardrailsMode:
+    """The event only blocks when both the global mode and the per-guardrail action are block."""
+    mode = get_policy_value(policy, 'mode', default=PolicyMode.BLOCK)
+    action = get_policy_value(feature_config, 'action', default=PolicyMode.BLOCK)
+    return GuardrailsMode.BLOCK if (mode == PolicyMode.BLOCK and action == PolicyMode.BLOCK) else GuardrailsMode.REPORT
+
+
+def build_ai_guardrails_scan_parameters(
```
omer-roth, line 321: Add a note to README.md that if someone is using AI guardrails - it will collect PII

omer-roth (changes requested):
added comments, please review

cycode/cli/apps/ai_guardrails/scan/handlers.py
Ilanlido: intended, we need to gather some information to identify the device

cycode/cli/apps/ai_guardrails/ides/copilot.py
Ilanlido: intentional yes, its only for session start

omer-roth (approved):

PR #520, opened 2026-08-16
CM-71014 copilot agent dialect

Ilanlido:
One hooks file is executed by more than one runtime: VS Code's chat runtime and
the Copilot agent runtime (Copilot CLI, and VS Code agent sessions). The agent
runtime keys its payload dialect off the CASE of the configured event key. Cycode
registered camelCase, so it answered with its own dialect (sessionId, no
hook_event_name), which matches_payload rejects - scans were skipped fail-open
with nothing surfaced beyond a -v debug line.

Register PascalCase event keys. Both runtimes then deliver the Claude-style
dialect already parsed here, verified live on Copilot CLI 1.0.75 and VS Code
1.133.

The tool vocabulary still differs per runtime, so accept both rather than switch:
VS Code reads files as read_file/filePath and names MCP tools mcp_<server>_<tool>;
the agent runtime uses Read/path and <server>-<tool>. Agent MCP servers are
declared in ~/.copilot/mcp-config.json, so that is read alongside VS Code's
mcp.json, and the server split is longest-match since server names may themselves
contain the separator. The agent reuses its read tool for directory listings with
an identical payload shape, so the path is stat-ed before it counts as a file read.

Also drop the timestamp-absent condition from ClaudeCode.matches_payload. Matching
on a field being absent is what broke Copilot scanning in the first place, and the
documented transcript_path is a sufficient positive test on its own.

Stale camelCase installs stay rejected; they are corrected by reinstalling hooks.

### DIFF
--- a/cycode/cli/apps/ai_guardrails/ides/copilot.py
+++ b/cycode/cli/apps/ai_guardrails/ides/copilot.py
@@ -326,7 +326,8 @@ def _read_file_path(tool_name: str, tool_input: object) -> Optional[str]:
     try:
         if not Path(raw_path).is_file():
             return None
-    except OSError:
+    except OSError as e:
+        logger.debug('Failed to stat read path, %s', {'path': raw_path}, exc_info=e)
         return None
     return raw_path
 
--- a/cycode/cli/apps/ai_guardrails/ides/copilot.py
+++ b/cycode/cli/apps/ai_guardrails/ides/copilot.py
@@ -1,17 +1,25 @@
-"""GitHub Copilot (VS Code extension) integration for AI guardrails.
+"""GitHub Copilot integration for AI guardrails.
 
 Hooks are installed in Copilot's native format to ``~/.copilot/hooks/cycode.json``
-(user scope) or ``<repo>/.github/hooks/cycode.json`` (repo scope). Both locations
-are also read by Copilot CLI and the Copilot cloud coding agent, but only the
-VS Code payload dialect is parsed here — CLI payloads (camelCase, no event name)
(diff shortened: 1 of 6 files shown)


cycode/cli/apps/ai_guardrails/ides/copilot.py
```diff
@@ -279,22 +308,56 @@ def _server_name_variants(server_name: str) -> set[str]:
...
+    """
+    if tool_name not in _READ_FILE_TOOLS or not isinstance(tool_input, dict):
+        return None
+
+    raw_path = next((tool_input[key] for key in _READ_PATH_KEYS if tool_input.get(key)), None)
+    if not isinstance(raw_path, str):
+        return None
+
+    try:
+        if not Path(raw_path).is_file():
```
omer-roth, line 327: is raw_path relative to the where the mcp is executed from or to the CLI's workdir?

If it's the first one, there's a chance is_file will fail because it's not an absolute path

cycode/cli/apps/ai_guardrails/ides/copilot.py
```diff
@@ -279,22 +308,56 @@ def _server_name_variants(server_name: str) -> set[str]:
...
+            return None
+    except OSError:
+        return None
```
omer-roth: Maybe worth adding a debug log to make troubleshooting easier here, just print the OSError

cycode/cli/apps/ai_guardrails/ides/copilot.py
Ilanlido: full path

omer-roth (approved):

PR #523, opened 2026-08-25
CM-68446: Render unmaintained package detections in SCA output

AradTraub:
## Summary

Renders unmaintained-package findings (OpenSSF Scorecard score <= 2) in local SCA scan output. Until now they arrived with an unrecognised policy id and fell through to a table titled "Unknown" with vulnerability-shaped columns.

- Dedicated **Unmaintained Packages** table with an OSSF Score column in place of CVE / Upgrade / License.
- Detail card gains the score and a link to the scorecard report.
- Package health is read from the nested `ossf` detection detail (`ossf.score`, `ossf.scorecard_report_url`); the flat `ossf_scorecard_score` field no longer exists.

**No new CLI flag.** Whether the policy runs is decided entirely by the tenant's policy state, resolved server-side in dependency-collector — the CLI neither sends nor needs a scan parameter for it. An earlier revision of this branch added a `--sca-scan unmaintained-packages` value; it was reverted once the gate moved to the policy, so this PR is rendering only.

## Testing

- `ruff check` + `ruff format --check` clean.
- `pytest`: 999 passed, 1 skipped, including new tests for the printer's title, its exact column set and order, and the score fallback when no scorecard is present.
- Verified end to end against a local backend: a CLI SCA scan on a tenant with the policy enabled prints the Unmaintained Packages table with a score of 1.5.


cycode/cli/printers/tables/sca_table_printer.py
```diff
@@ -120,6 +130,9 @@ def _enrich_table_with_values(table: Table, detection: Detection) -> None:
         table.add_cell(CVE_COLUMNS, detection_details.get('vulnerability_id'))
         table.add_cell(LICENSE_COLUMN, detection_details.get('license'))
 
+        ossf_score = get_ossf_score(detection_details)
+        table.add_cell(OSSF_SCORE_COLUMN, 'N/A' if ossf_score is None else str(ossf_score))
```
omer-roth: Is this the overall ossf score or the maintained check's score?

cycode/cli/apps/scan/scan_parameters.py
```diff
@@ -16,6 +16,7 @@ def _get_default_scan_parameters(ctx: typer.Context) -> dict:
         'report': ctx.obj.get('report'),
         'package_vulnerabilities': ctx.obj.get('package-vulnerabilities'),
         'license_compliance': ctx.obj.get('license-compliance'),
+        'maintainability': ctx.obj.get('unmaintained-packages'),
```
omer-roth: overall might be worth checking:
What happens if the policy is off and the user specified they want to scan unmaintained-packages specifically?

AradTraub: they will get 0 detections. as the policy is close and this the maintainability will be align, there will be no detection found as the rest of the detection will be filtered out and maintainability will skipped due to the fact that the policy is close

cycode/cli/printers/tables/sca_table_printer.py
AradTraub: The overall score.

cycode/cli/apps/scan/scan_parameters.py
```diff
@@ -16,6 +16,7 @@ def _get_default_scan_parameters(ctx: typer.Context) -> dict:
         'report': ctx.obj.get('report'),
         'package_vulnerabilities': ctx.obj.get('package-vulnerabilities'),
         'license_compliance': ctx.obj.get('license-compliance'),
+        'maintainability': ctx.obj.get('unmaintained-packages', False),
```
avishaiamiel: Is there a reason that you added default value?
If not, let's be consistent

cycode/cli/printers/tables/sca_table_printer.py
```diff
@@ -120,6 +130,9 @@ def _enrich_table_with_values(table: Table, detection: Detection) -> None:
         table.add_cell(CVE_COLUMNS, detection_details.get('vulnerability_id'))
         table.add_cell(LICENSE_COLUMN, detection_details.get('license'))
 
+        maintained_score = get_maintained_score(detection_details)
+        table.add_cell(MAINTAINED_SCORE_COLUMN, 'N/A' if maintained_score is None else str(maintained_score))
```
omer-roth: Please put this being policy gate (score is irrelevant for other policies

cycode/cli/printers/rich_printer.py
```diff
@@ -97,7 +98,13 @@ def __add_sca_scan_related_rows(details_table: Table, detection: 'Detection') ->
...
         details_table.add_row('Dependency path', dependency_path or 'N/A')
 
-        if not detection.has_alert:
+        if detection.detection_type_id == consts.UNMAINTAINED_PACKAGE_POLICY_ID:
+            maintained_score = get_maintained_score(detection_details)
+            ossf_score = get_ossf_score(detection_details)
+            details_table.add_row('Maintained score', 'N/A' if maintained_score is None else str(maintained_score))
+            details_table.add_row('OSSF Scorecard score', 'N/A' if ossf_score is None else str(ossf_score))
+            details_table.add_row('Scorecard report', get_ossf_report_url(detection_details) or 'N/A')
+        elif not detection.has_alert:
```
omer-roth: should this actually be conditional?

cycode/cli/printers/text_printer.py
```diff
@@ -84,7 +85,17 @@ def __get_intermediate_summary_lines(self, detection: 'Detection') -> list[str]:
...
+            maintained_score = get_maintained_score(detection.detection_details)
+            ossf_score = get_ossf_score(detection.detection_details)
+            maintained = 'N/A' if maintained_score is None else maintained_score
+            score = 'N/A' if ossf_score is None else ossf_score
+            report_url = get_ossf_report_url(detection.detection_details) or 'N/A'
+
+            summary_lines.append(f'Maintained score: [cyan]{maintained}[/]\n')
+            summary_lines.append(f'OSSF Scorecard score: [cyan]{score}[/]\n')
+            summary_lines.append(f'Scorecard report: [cyan]{report_url}[/]\n')
+        elif detection.has_alert:
```
omer-roth: should this actually be conditional?

omer-roth (changes requested):
Added some comments

cycode/cli/printers/rich_printer.py
AradTraub: Yes — deliberately, and it is the bug this PR fixes.

`Detection.has_alert` documents the old assumption in its own docstring (`cyclient/models.py:37-41`): *"For example, for SCA, it means that the detection is a package vulnerability. Otherwise, it is a license."* That binary held while SCA had two policies.

Unmaintained packages are a third, and they carry neither an `alert` nor a `license` — so under the previous code every unmaintained finding rendered a `License` row with no value. Making it an `elif` restores one branch per policy.

---
🤖 Replied with Claude Code

cycode/cli/printers/text_printer.py
AradTraub: Same as the rich printer — yes, deliberately.

Here the old code went `if detection.has_alert: … else: License`, so an unmaintained detection (no `alert`) fell into the `else` and printed `License: N/A` on every finding. That was the concrete bug; the `elif` gives each of the three SCA policies its own branch.

---
🤖 Replied with Claude Code

omer-roth (approved):

PR #538, opened 2026-09-02
CM-71972: Collect Claude Code skills in the Guardrails session sweep

Altruistus:
## TL;DR

A developer installs a skill from a marketplace into `~/.claude/skills/`, never commits it, and Cycode's Skills inventory shows nothing. The session-start hook already sweeps every IDE for MCP configs; this adds skills to the same sweep and the same report, so an installed skill becomes visible without anyone having to commit it.

## Why

Skills inventory has only ever seen a `SKILL.md` that reached a repository. The instructions and scripts a marketplace skill carries have been invisible — worst of all for people with no SCM footprint at all. The backend can already inventory a reported skill; it just needs the CLI to report one.

## Reviewer notes

- **Skills get their own collector rather than riding in `config_files[]`.** The backend parses every entry there as an MCP server map, and a skill is a directory of Markdown, not a normalizable JSON config.
- **Plugin skills are the exception and hang off the plugin entry**, because that is what carries the marketplace, plugin and version a marketplace-installed skill came from. The plugin readers already hold the resolved directory, so the scan is shared and all three IDEs with a plugin system get it — the `skills/` layout belongs to the plugin format, not to any one IDE.
- **Raw content is sent, not parsed frontmatter.** Parsing belongs to the backend, which is also the only option for the device connectors that read these files off endpoints and can return nothing but file bytes.
- **Two caps that the MCP collectors did not need.** A `SKILL.md` body is unbounded prose and the number of installed skills is unbounded too. One request carries the whole session context, so an oversized skill would otherwise cost the device its MCP inventory as well.
- **User-scope collection is Claude Code only, deliberately.** It is the only tool with a `SKILL.md` convention of its own today — Cursor's `.mdc` rules and `CLAUDE.md` are rule files, a different category. The hook is generic (`IDE.get_skills` defaults to empty), so adding another IDE is two lines once its skills directory is confirmed rather than guessed.
- **A test-isolation bug surfaced on the first run and is fixed here.** The existing session-context tests read the developer's real `~/.claude/skills`, so their assertions depended on whose machine ran them. An autouse fixture now pins the sweep.
- Project-scope skills (`<repo>/.claude/skills/`) are out of scope: there is no repo-scope collection in this app at all, which is structural plumbing rather than a missing path.

[CM-71972](https://cycode.atlassian.net/browse/CM-71972)

[CM-71972]: https://cycode.atlassian.net/browse/CM-71972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cycode/cli/apps/ai_guardrails/ides/claude_code.py
+++ b/cycode/cli/apps/ai_guardrails/ides/claude_code.py
@@ -13,6 +13,7 @@ from cycode.cli.apps.ai_guardrails.ides._plugin_utils import (
     resolve_cached_plugin_dir,
     walk_enabled_plugins,
 )
+from cycode.cli.apps.ai_guardrails.ides._skill_utils import walk_skill_dirs
 from cycode.cli.apps.ai_guardrails.ides.base import IDE, DecisionAction, HookDecision
 from cycode.cli.apps.ai_guardrails.scan.payload import AIHookPayload
 from cycode.cli.apps.ai_guardrails.scan.types import AiHookEventType
@@ -33,6 +34,10 @@ _HOOK_EVENTS = ['UserPromptSubmit', 'PreToolUse:Read', 'PreToolUse:mcp']
 
 _CLAUDE_CONFIG_PATH = Path.home() / '.claude.json'
 _CLAUDE_SETTINGS_PATH = Path.home() / '.claude' / 'settings.json'
+_CLAUDE_SKILLS_DIR = Path.home() / '.claude' / 'skills'
+
+# Claude hardcodes a plugin's skills under this subdirectory, the same way it hardcodes ".mcp.json".
+_PLUGIN_SKILLS_SUBDIR = 'skills'
 
 _SCAN_COMMAND = f'{CYCODE_SCAN_PROMPT_COMMAND} --ide claude-code'
 _SESSION_START_COMMAND = f'{CYCODE_SESSION_START_COMMAND} --ide claude-code'
@@ -187,16 +192,23 @@ def _resolve_marketplace_path(marketplace: dict) -> Optional[Path]:
 
 
 def _read_claude_plugin(plugin_dir: Path) -> tuple[dict, dict]:
-    """Read one Claude Code plugin's manifest + MCP servers.
+    """Read one Claude Code plugin's manifest, MCP servers and skills.
 
     Claude hardcodes the MCP file at ``<plugin_dir>/.mcp.json`` and always
-    wraps it as ``{"mcpServers": {...}}``.
+    wraps it as ``{"mcpServers": {...}}``, and a plugin's skills at
+    ``<plugin_dir>/skills/<name>/SKILL.md``.
     """
     manifest = load_plugin_json(plugin_dir / '.claude-plugin' / 'plugin.json') or {}
     entry: dict = {}
     for field in ('name', 'version', 'description'):
         if field in manifest:
             entry[field] = manifest[field]
+
+    # Attached to the plugin entry rather than the top-level skills list so the backend keeps the
+    # plugin provenance (marketplace, plugin, version) that a marketplace-installed skill has.
+    skill_files = walk_skill_dirs(plugin_dir / _PLUGIN_SKILLS_SUBDIR)
+    if skill_files:
+        entry['skill_files'] = skill_files
 
     mcp_config_path = plugin_dir / '.mcp.json'
     mcp_config = load_plugin_json(mcp_config_path) or {}
@@ -387,3 +399,6 @@ class ClaudeCode(IDE):
         enriched_plugins = resolve_plugins(settings) if settings else {}
 
         return global_config_file, enriched_plugins
+
+    def get_skills(self) -> list[dict]:
+        return walk_skill_dirs(_CLAUDE_SKILLS_DIR)
--- a/cycode/cli/apps/ai_guardrails/ides/claude_code.py
+++ b/cycode/cli/apps/ai_guardrails/ides/claude_code.py
@@ -13,7 +13,7 @@ from cycode.cli.apps.ai_guardrails.ides._plugin_utils import (
     resolve_cached_plugin_dir,
     walk_enabled_plugins,
 )
-from cycode.cli.apps.ai_guardrails.ides._skill_utils import walk_skill_dirs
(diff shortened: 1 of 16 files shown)


cycode/cli/apps/ai_guardrails/ides/claude_code.py
```diff
@@ -387,3 +404,6 @@ def get_session_context(self) -> tuple[Optional[dict], dict]:
         enriched_plugins = resolve_plugins(settings) if settings else {}
 
         return global_config_file, enriched_plugins
+
+    def get_skills(self) -> list[dict]:
```
Ilanlido, line 408: What about supporting other agents like codex etc.

cycode/cli/apps/ai_guardrails/ides/copilot.py
```diff
@@ -178,6 +179,12 @@ def _read_copilot_plugin(plugin_dir: Path) -> tuple[dict, dict]:
         if field in manifest:
             entry[field] = manifest[field]
 
+    # Same plugin-format layout as every other IDE's plugins, so a plugin shipping skills is
+    # inventoried whichever IDE loaded it.
+    skill_files = walk_plugin_skills(plugin_dir)
```
Ilanlido, line 184: What about plugin skills for cursor?

tests/cli/commands/ai_guardrails/test_session_start_command.py
```diff
@@ -31,6 +31,17 @@ def _isolated_session_context_cache(tmp_path: Path, monkeypatch: pytest.MonkeyPa
     monkeypatch.setattr(_session_start_mod, '_session_context_cache_path', lambda: tmp_path / '.session-context-cache')
 
 
+@pytest.fixture(autouse=True)
+def _no_local_skills(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Keep the skills sweep away from the developer's real ~/.claude/skills.
+
+    Unlike the MCP collectors, which these tests stub per IDE, the skills sweep walks the
+    filesystem directly - so without this every assertion would depend on whoever ran it.
```
Ilanlido, line 39: Really? Aren't the units using fakefs?

tests/cli/commands/ai_guardrails/test_session_start_command.py
```diff
@@ -590,3 +605,93 @@ def test_unauthenticated_skips_session_init(
...
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """User-scope skills ride alongside the MCP inventory in the same report."""
+    mock_get_auth.return_value = MagicMock(tenant_id='tenant-1')
+    mock_ai_client = MagicMock()
+    mock_get_client.return_value = mock_ai_client
+    mock_collect.return_value = ({}, {})
+    skill_path = '/home/u/.claude/skills/dummy-skill/SKILL.md'
+    skills = [{'path': skill_path, 'content': '---\nname: dummy-skill\n---\nBody.\n'}]
+    monkeypatch.setattr(_session_start_mod, 'collect_all_skills', lambda: skills)
```
Ilanlido, line 632: Why is this mocked

cycode/cli/apps/ai_guardrails/ides/claude_code.py
Altruistus: Done in 88d4018 — Codex, Copilot and Cursor now each have `get_skills()` reading their own skills directory (`~/.codex/skills`, `~/.copilot/skills`, `~/.cursor/skills`), honoring `$CODEX_HOME`/`$COPILOT_HOME` like the rest of those integrations.

Nothing else needed changing: `collect_all_skills()` already sweeps every registered IDE and dedupes by path, and `walk_skill_dirs` returns an empty list for a directory that isn't there — so an IDE that has no skills just contributes none. Added the same two tests per IDE that Claude Code already had.

cycode/cli/apps/ai_guardrails/ides/copilot.py
Altruistus: Cursor has no plugin system in this integration to hang them off — `Cursor.get_session_context()` reads `~/.cursor/mcp.json` and returns `{}` for plugins, with no `walk_enabled_plugins` call, because Cursor has no `<plugin>@<marketplace>` layout like Claude Code, Codex and Copilot do. So there is no plugin directory to scan for `skills/`.

If Cursor grows one, the extension point is the same three lines the other three use inside their `_read_*_plugin`, and `walk_plugin_skills` needs no change.

Its **user-scope** skills are covered as of 88d4018 (`~/.cursor/skills`) — see the other thread.

tests/cli/commands/ai_guardrails/test_session_start_command.py
Altruistus: Good catch, and you're right that the fixture was the wrong shape — but not for the reason you'd expect: **this module doesn't use pyfakefs at all.** No `fs` fixture anywhere in it; it isolates by patching the per-IDE collectors instead. So the skills sweep, which walks the filesystem directly rather than going through a collector, really was reading the real `~/.claude/skills` of whoever ran the tests. I hit it on my own machine — the assertions picked up my seven actual skills.

Rather than stub an application function in an autouse fixture, 88d4018 isolates at the filesystem level instead: `_isolated_home` points `Path.home()` at a scratch dir, so nothing home-relative can reach real files. That also let both skill tests drop their mocks (see the other thread).

tests/cli/commands/ai_guardrails/test_session_start_command.py
Altruistus: Fair — it shouldn't be. It was only mocked because the autouse fixture stubbed `collect_all_skills` globally, so a test that cared about skills had to override the stub. Circular, and it meant neither skill test actually exercised the sweep.

Fixed in 88d4018. With `Path.home()` redirected to a scratch dir, both tests write real `SKILL.md` files and let the real `collect_all_skills()` find them:

- `test_reports_skill_files` creates the file and asserts the reported `skill_files` — so it now covers `walk_skill_dirs` and the payload wiring, not just the wiring.
- `test_editing_a_skill_re_reports` rewrites the file's content between runs, so the dedup-digest assertion goes through a genuine re-read.

Nothing about skills is mocked in that module any more.

Ilanlido (approved):

omer-roth (approved):

PR #541, opened 2026-09-07
CM-68944: Fetch AI guardrail config from the platform 

Ilanlido:
Enforcement config (per-agent Off/Report/Block modes and sensitive-path globs) is now platform-owned. session-start fetches the tenant's resolved config into ~/.cycode/ai-guardrails-config.json (TTL- and tenant-gated); scans only read the cache and overlay it onto the local policy, so the hot path never touches the network. No cache means built-in Report defaults, always synchronous.

- An event whose guardrails are all Off exits before any scan, event or auth.
- The two FileRead guardrails (content scan, sensitive path) keep independent modes via a per-check action.
- ai-guardrails.yaml is reduced to operational knobs (timeout, max bytes, fail_open); enforcement keys older CLIs wrote are stripped, and --mode on install is deprecated.
- Atomic write and corrupt-file quarantine are shared helpers in path_utils.


cycode/cli/apps/ai_guardrails/scan/consts.py
Ilanlido: These fields shouldnt have meaning now too since its in platform no?

Ilanlido: Same

cycode/cli/apps/ai_guardrails/scan/guardrail_config.py
```diff
@@ -0,0 +1,170 @@
...
+from cycode.cli.consts import CYCODE_CONFIGURATION_DIRECTORY
+from cycode.cli.utils.path_utils import atomic_write_text, quarantine_corrupt_file
+from cycode.logger import get_logger
+
+logger = get_logger('AI Guardrails')
+
+GUARDRAILS_CONFIG_FILE_NAME = 'ai-guardrails-config.json'
+
+# The matrix cell values; Report and Block share GuardrailsMode's spelling.
+MODE_OFF = 'off'
```
Ilanlido: add this value to the enum?

cycode/cli/utils/path_utils.py
```diff
@@ -86,6 +87,32 @@ def get_file_content(file_path: Union[str, 'PathLike']) -> Optional[AnyStr]:
         logger.warn('Permission denied to read the file: %s', file_path)
 
 
+def atomic_write_text(filename: str, content: str) -> None:
+    """Write via a temp file + rename so concurrent CLI processes never read a torn file."""
+    directory = os.path.dirname(filename)
+    file_descriptor, temp_filename = tempfile.mkstemp(dir=directory, prefix=f'.{os.path.basename(filename)}.')
```
MaorDavidzon: Don't use temp files. Write it to .cycode and delete it.
This is part of what caused us the issue with virustotal

cycode/cli/apps/ai_guardrails/scan/handlers.py
```diff
@@ -109,7 +112,9 @@ def handle_before_read_file(ctx: typer.Context, payload: AIHookPayload, policy:
         return HookDecision.allow(AiHookEventType.FILE_READ)
 
     file_path = payload.file_path or ''
-    effective_mode = get_effective_mode(policy, file_read_config)
+    # Two guardrails share this event, each with its own mode: the path match and the content scan.
```
Ilanlido: no need for the comment

cycode/cli/apps/ai_guardrails/scan/policy.py
```diff
@@ -83,21 +115,22 @@ def load_policy(workspace_root: Optional[str] = None) -> dict:
     policy = load_defaults()
 
     # Merge machine-wide config (admin/MDM-provisioned) - overrides defaults, below user/repo.
-    machine_config = load_yaml_file(get_machine_policy_path())
+    machine_policy_path = get_machine_policy_path()
```
Ilanlido, line 117: tbh - if we have these settings from platform - maybe we can delete the machine settings? what is its use now?

cycode/cli/apps/ai_guardrails/scan/policy.py
```diff
@@ -83,21 +115,22 @@ def load_policy(workspace_root: Optional[str] = None) -> dict:
...
+    machine_config = load_yaml_file(machine_policy_path)
     if machine_config:
-        policy = deep_merge(policy, machine_config)
+        policy = deep_merge(policy, strip_platform_managed_keys(machine_config, str(machine_policy_path)))
```
Ilanlido: im not sure about this strip_platform_managed_keys feels a bit complex - maybe the policy should be just the operational knobs and then just merge platform settings on top?

cycode/cli/apps/ai_guardrails/scan/consts.py
Altruistus: Right, both were dead. `action` is overwritten by `apply_platform_config` on every scan, and `enabled` could never be false: a local file's `enabled` is dropped on load and the platform never sets it.

It was also worse than dead - `create_policy_file` wrote those keys into the user's `ai-guardrails.yaml` on install, so we were shipping a file full of settings that do nothing.

`DEFAULT_POLICY` is now operational knobs only (`version`, `fail_open`, `secrets`), and the dead `enabled` gates are out of the three handlers.

Altruistus: Same, with one exception: `deny_globs` is still needed as the cold-start fallback. `apply_platform_config` overwrites it from the platform, but a scan that runs before session-start has cached anything has no other list to match against. It moved out of `DEFAULT_POLICY` into `DEFAULT_SENSITIVE_PATH_GLOBS` so it reads as what it is.

`scan_content` stays as a runtime key (the platform sets it from the SecretsInFile cell) - it just no longer needs a default. `mcp.scan_arguments` is gone entirely; see the `policy.py` thread, it was an open hole rather than a dead key.

cycode/cli/apps/ai_guardrails/scan/guardrail_config.py
Altruistus: Added, but as its own enum rather than a member on `GuardrailsMode`.

`GuardrailsMode` is also the type of the `ai-guardrails install --mode` option and the mode we report on the scan event, so an `OFF` member would surface `--mode off` as a valid install choice - which it isn't, Off is a platform-only state.

So: new `GuardrailCellMode` (`off`/`report`/`block`, the last two reusing `GuardrailsMode`'s spellings) in `ai_guardrails/consts.py`, and `MODE_OFF`/`MODE_REPORT`/`MODE_BLOCK` are gone. `mode_for` still returns a plain lowercased string so an unknown value from the server can't raise; the comparisons still work because it's a `str` enum.

cycode/cli/apps/ai_guardrails/scan/handlers.py
Altruistus: Removed.

cycode/cli/apps/ai_guardrails/scan/policy.py
Altruistus: Kept, but it's a fair question.

After this change the machine file can only carry `fail_open`, `secrets.timeout_ms`, `secrets.max_bytes` and `version` - no enforcement at all. That's still the only layer an admin can write on a managed fleet: MDM can drop `/Library/Application Support/Cycode/ai-guardrails.yaml`, it can't write into every user's `~/.cycode`.

If we decide those knobs shouldn't be fleet-tunable either, then the whole local policy file goes, not just the machine layer. Happy to do that as a follow-up, but it's a behaviour change for the MDM installs so I didn't fold it into this PR.

cycode/cli/apps/ai_guardrails/scan/policy.py
Altruistus: Agreed - changed to exactly that shape. `load_policy` now whitelists what a local file may contribute (`_LOCAL_POLICY_KEYS` = `version`, `fail_open`, `secrets`) and ignores everything else, then `apply_platform_config` overlays the platform config on top.

It also closes a hole: the strip-list didn't include `mcp.scan_arguments`, so a repo-level `.cycode/ai-guardrails.yaml` with `mcp: {scan_arguments: false}` switched off MCP argument scanning entirely - a local file disabling a platform guardrail. A whitelist can't have that class of bug, and the repo-config test now asserts that nothing a local file says about enforcement survives.

`strip_platform_managed_keys` survives only on the write path (`create_policy_file`), where it just drops the four platform sections. We can't whitelist there - it would delete whatever unrelated keys the user has in their own file.

cycode/cli/apps/ai_guardrails/scan/consts.py
```diff
@@ -1,48 +1,37 @@
...
+DEFAULT_SENSITIVE_PATH_GLOBS = [
+    '.env',
+    '.env.*',
+    '*.pem',
+    '*.p12',
+    '*.key',
+    '.aws/**',
+    '.ssh/**',
+    '*kubeconfig*',
+    '.npmrc',
```
omer-roth, line 24: (Optional)
`.npmrc` is the nodeJS variant for providing credentials to private registries.
We have different files for python, java, pnpm (nodejs too) and every other ecosystem.

Would it be easier to have an include list instead of exclude list?

omer-roth (approved):

cycode/cli/apps/ai_guardrails/scan/consts.py
Altruistus: You're right that it's Node-only, and it's worse than that — I ran the list against the real matcher:

```
ALLOWED  ~/.kube/config          <- canonical kubeconfig path
ALLOWED  ~/.pypirc               <- Python
ALLOWED  ~/.m2/settings.xml      <- Maven
ALLOWED  ~/.docker/config.json
ALLOWED  ~/.git-credentials
ALLOWED  ~/.gem/credentials
ALLOWED  ~/.cargo/credentials.toml
```

`.kube/config` is the one that surprised me: `*kubeconfig*` only matches a file with "kubeconfig" in the name, so the default `~/.kube/config` sails straight through while `~/kubeconfig` is blocked.

On include vs exclude though — an allowlist can't work for this guardrail. It fires on *every file the agent reads*, which for a coding agent is the whole repo plus whatever it wanders into. An allowlist would have to enumerate everything readable and deny by default, so the agent would be blocked from reading source code. Deny-listing credential files is the only shape that leaves the tool usable; the cost is exactly what you spotted, that the list needs maintaining per ecosystem.

What I'd rather not do is fix it *here*. After this PR that constant is only the cold-start fallback — the list that actually enforces comes from the platform (`GuardrailCatalog.DefaultSensitiveGlobs` in ai-security-manager, merged in !114), and `apply_platform_config` overwrites `deny_globs` on every scan once session-start has cached it. The two are documented as mirroring each other, so changing this copy alone would diverge them without changing what any tenant actually enforces.

So the real fix is in ai-security-manager, mirrored back here. Worth its own ticket, partly because "which credential files, across which ecosystems" is security content someone should own deliberately rather than me guessing in a review reply.

One upside in the meantime: this PR is what makes that list tenant-editable. An admin can add `.pypirc` or `.kube/config` to their own sensitive-path list from the platform today, without a CLI release.

PR #546, opened 2026-09-14
CM-72432: Retry token mint when a concurrent login is refused

Ilanlido:
AI guardrails runs one CLI process per hook, and every process shares one cached token with a single expiry stamp, so a burst of them mints at the same instant. The identity provider's brute-force protection refuses same-user logins that land within milliseconds of each other, so all but the first are rejected. The refused scans fail open and record a scan failure, leaving the interaction unscanned.

On a 401 during refresh, wait briefly and prefer the token the winning process has just persisted, minting again only if nobody won. The wait is randomized so the refused processes do not collide a second time.


cycode/cyclient/base_token_auth_client.py
```diff
@@ -49,7 +57,20 @@ def refresh_access_token_if_needed(self) -> None:
         self._load_token_from_disk()
         if self._has_valid_token():
             return
-        self.refresh_access_token()
+
+        try:
+            self.refresh_access_token()
+        except HttpUnauthorizedError:
```
omer-roth, line 63: it's ok to continue if other exceptions are thrown? time outs, forbidden, etc.?

omer-roth (approved):

